PR #5, opened 2020-05-22
KCON-1: search for conference focus before initiating 'dial'

jamesaimonetti:


jamesaimonetti:
Requires:
- https://github.com/2600hz/kazoo-ecallmgr/pull/6
- https://github.com/2600hz/kazoo-core/pull/12

PR #8, opened 2020-05-29
KZOO-65: include reboot flag when check-sync event sent to devices

jamesaimonetti:


jamesaimonetti:
Needs:
- https://github.com/2600hz/kazoo-core/pull/17
- https://github.com/2600hz/kazoo5/pull/10

PR #11, opened 2020-06-03
Handle the error message if its returned from the validation function

bradfordben:
Requires https://github.com/2600hz/kazoo-core/pull/19

Handle the new added return type of `{'system_error', {Error, Message}}`


jamesaimonetti:
@bradfordben it seems that the kzd modules' specs aren't update to date with this error tuple construction. Might need another PR to address that shortcoming.

bradfordben:
> @bradfordben it seems that the kzd modules' specs aren't update to date with this error tuple construction. Might need another PR to address that shortcoming.

Its in the required PR linked in the description as its in core.

jamesaimonetti:
@bradfordben you're right, missed that.

bradfordben:
@jamesaimonetti No problem, thanks for reviewing 👍

PR #23, opened 2020-06-23
[Master] PROD-145: Add read-only scope to Kazoo authentication

mk1s:
# PROD-145: Add read-only scope to Kazoo authentication

This PR introduces several modules for the definition and enforcement of authentication scopes within the `kazoo_auth` workflow.

Currently, API restrictions are enforced using token restrictions which define resource and verb privileges for crossbar to use during request validation.  The provided backend modules are designed to be used with token restrictions and preserve any existing permission sets.

Scope enforcement is disabled by default and merges scopes with token restrictions in flight and does not modify the token restrictions definitions.

#### Related PRs
* [kazoo-core #25](https://github.com/2600hz/kazoo-core/pull/25)
* [kazoo-crossbar #23](https://github.com/2600hz/kazoo-crossbar/pull/23)

#### Enabling Scope Enforcement

To enable scope enforcement, set `iam.enforce_scopes` to **true** in the `system_config/crossbar.auth`.  

For additional configuration details, consult the **scopes.md** in `kazoo_auth` docs.

#### Database Maintenance
On startup, the `kazoo_auth_app` will create the `iam` database, load views, as well as the default scopes from their fixture files. 

Additionally, the new schema `scopes.json` should be automatically loaded on startup by crossbar.

## <a href="#usage">Usage</a>
Scopes and scope filtering can be requested during the authentication process.

**Example:** Read Only Scope
```sh
curl --request PUT \
  --url http://{CROSSBAR_URI}:8000/v2/user_auth \
  --header 'content-type: application/json' \
  --data '{
	"data": {
		"credentials": {CREDENTIAL_MD5},
		"method": "md5",
		"account_name": "A",
		"scope": "read_only"
	}
}'
```

## <a href="#backend">Kazoo Scope Backend</a>
A new OTP behaviour, `kz_scope_backend`, has been added which allows you to create a custom mapping module for scopes and their behaviors.

At this time the only available backend is the `cb_tr_scope_backend` which attempts to map API related scopes to the `token_restrictions` database.

### cb\_tr\_scope\_backend
This backend utilizes crossbar token restrictions and attempts to merge any scopes defined in the JWT with the restrictions defined for the requesting owner.

### Available Scopes
Default scopes and their claims are automatically added to the JWT during authentication in order to maintain backwards compatibility.

#### System Scopes
* identity<sup>1</sup>
* role<sup>1</sup>
* read_only
* auth_only

##### OpenID Connect Scopes
* openid<sup>1</sup>
* profile
* email

<sup>1</sup> _indicates a default scope_


jamesaimonetti:
It appears that at least `api_auth` responses are now arrays instead of objects:

`{"data":[{"account_id":"...",...}], ...}` is what I see in the kazoo-properly tests and is causing issues. Reverting Crossbar back to master undoes the array'd result.

PR #32, opened 2020-08-10
view-abstract: use kz_view cursor

icehess:
Move all pagination and chunking to the new `kz_view` and `kz_view_cursor` modules from core.


src/crossbar_view.erl
```diff
@@ -236,7 +236,7 @@ build_time_range_keys(_Context, {'ok', CursorOptions}) ->
             {'ok', props:set_values(Prop, props:delete('bookmark', CursorOptions))}
     end;
 build_time_range_keys(_, {'error', Context}) ->
-    Context.
+    {'error', Context}.
```
icehess: same for line build_range_keys at line 220

PR #36, opened 2020-08-13
PISTON-1112: fix validation of agent pause timeout/time-limit

danielfinke:
Crossbar component of https://github.com/kazoo-community/kazoo-acdc/pull/31

- schema validates system_config.acdc default_agent_pause_timeout for non-negative integers or the "infinity" special string
- refactored out ?DEFAULT_AGENT_PAUSE_TIMEOUT into shared header
- fixed acdc_maintenance expecting integer, despite possibility of <<"infinity">> value
- moved Time-Limit validation in kapi_acdc_agent into PAUSE_* macros rather than AGENT_*


priv/couchdb/schemas/system_config.acdc.json
```diff
@@ -39,7 +39,12 @@
         "default_agent_pause_timeout": {
             "default": 600,
             "description": "acdc default agent pause timeout",
-            "type": "integer"
+            "minimum": 0,
+            "pattern": "infinity",
+            "type": [
```
jamesaimonetti: I think OAS doesn't like `type` to be an array. Also I would suggest making it an integer (which doesn't preclude it from being a string in the DB but API manipulation will convert it to integer).

danielfinke: @jamesaimonetti it is, however, a valid value for it to also be "infinity". The API should support setting that particular string. Any ideas how to address this?

jamesaimonetti: Sorry, how is agent pause = 'infinity' a useful thing? :)

I don't have a good idea here tbh. @icehess thoughts on schema?

danielfinke: @jamesaimonetti a number of our call centre users have desk phones with BLF showing their agent pause state and use that button to toggle pause. They also use the separate "pause aliases" feature (can't recall if we PR'd this one in 4.x) to put themselves on pause for various reasons/aliases such as "Lunch" or "Coaching". All of these have wildly different and inconsistent time frames, so these customers explicitly perform a resume when they no longer want to be on pause.

icehess: Even not considering OAS compatibility I think this is not even a valid schema. mixing both integer and string with all their property like minimum or pattern. Even if valid it is confusing.

You can use `oneOf` like this:

```json
        "default_agent_pause_timeout": {
            "description": "acdc default agent pause timeout",
            "oneOf": [
                {
                    "default": 600,
                    "minimum": 0,
                    "type": "integer"
                },
                {
                    "enum": [
                        "infinity"
                    ],
                    "type": "string"
                }
            ]
        },
```

icehess (changes requested):

PR #38, opened 2020-08-27
KCRO-32: validation no longer uses crossbar_doc so the context db_doc…

k-anderson:
… is not set and billing relies on that


k-anderson:
4.3 PR is here https://github.com/2600hz/kazoo/pull/6625

PR #44, opened 2020-09-03
[master] SUPP-16: load_parents sets an empty object for account

icehess:
It is not clear what is the difference between `/accounts/:id/parents` and
`/accounts/:id/tree`. But the are doing same thing in different ways.

As it turns out `load_parents` is loading all result of `accounts/listing_by_id`
and then uses `kz_json:find_value/3` to grab the parent's account info.

But it is constantly failing in a specific server. I can't reproduce the exact
problem in any environment that I test it.

For example it should return:

```json
[{"id":"{ACCOUNT_1}","name":"{ACCOUNT1_NAME}"},{"id":"{ACCOUNT_2}","name":"{ACCOUNT2_NAME}"}]
```

But it returns instead:

```json
[{"id":"{ACCOUNT_1}","name":"{ACCOUNT1_NAME}"},{}]
```

Since those `/tree` and `/parents` seems returning same thing
and `/tree` implementation is way more efficient and stable than `/parents`
I change it to use that.

4.3: https://github.com/2600hz/kazoo/pull/6627


src/modules/cb_accounts.erl
```diff
@@ -867,10 +862,22 @@ load_siblings_results(_AccountId, Context, [JObj|_]) ->
...
+    case cb_context:resp_status(Context1) of
+        'success' -> load_account_tree(Context1);
+        _Else -> Context1
+    end.
+
 -spec load_account_tree(cb_context:context()) -> cb_context:context().
 load_account_tree(Context) ->
     Tree = get_authorized_account_tree(Context),
-    case kz_datamgr:open_cache_docs(?KZ_ACCOUNTS_DB, Tree) of
+    case kz_datamgr:open_docs(?KZ_ACCOUNTS_DB, Tree) of
```
jamesaimonetti: Are you seeing stale results in the cache? Why not use it?

jamesaimonetti:
@icehess see https://github.com/2600hz/kazoo-properly/pull/22

Unable to recreate it as well; guessing the one node you have seen this has bad data or something else amiss. Not opposed to the changes here but we should probably put a deprecation period on `parents` or make it follow `tree`'s code paths instead.

Thoughts?

PR #47, opened 2020-09-17
Adding additionalProperties to all account config

bradfordben:
Adding additionalProperties to all account config schemas. This is required for PR https://github.com/2600hz/kazoo-core/pull/77 so that the config api will filter out any additional properties not defined in the account config schema

4.3 PR: https://github.com/2600hz/kazoo/pull/6631


lazedo (changes requested):
not merging this

bradfordben:
Rebased off master.

PR #50, opened 2020-09-20
PROD-330: sparse fieldsets, array filter and full_docs features

icehess:
This will adds the ability to request sparse field set of the response and a ne wset of crossbar filters to filter array values.

* Allows users to request an endpoint to return specified `fields` in response using `fields` query string. This is limited to public fields of the document. 
* Add `filter_array_intersect_all_KEY`, `filter_array_intersect_any_KEY`, `filter_array_intersect_none_KEY` to filter docs by value of array `KEY` in document.
* Add new Crossbar filter `filter_any_KEY=["v1","v2"]` (and its negate version `filter_none_KEY`) query string to filter docs if `KEY` fields in the document is matching any of the elelments of this filter.

There are two options available that allows modules to control sparse fieldset features:

1. `no_fields`: modules can set this to turn off this feature in crossbar_view.
2. `field_key`: modules can set this to change the result key (doc, value, id, key or special `filtermap` which will use result from filtermap) that is being used to requested fields from. Default value is `doc`.

The `fields` and `field_key` options will be passed to `kz_view`.

If request has `fields` and `no_fields` is false, for value of `field_key`:

1. If `field_key` is `filtermap`, crossbar will create the filtermap function (using mapper and filters if any)
2. For other `field_key` values, if request has some qs filters, a qs filter will be generate as filtermap function
3. If none of above, then there will be no filtermap generated.

## Update

Also adding ability so client can request the full public fields of documents when listing.

This feature is opt-in, can be enabled for all modules using `allow_fetch_full_docs`
config or per modules `allowed_modules_fetch_full_docs.{CB_MODULE_NAME}`.

`fields` query string option precedents over `full_docs` is both have been specified.

## Required Core PR:

~parse array in query string: https://github.com/2600hz/kazoo-core/pull/75~ We don't use this anymore. The array should be an encoded JSON array, e.g. `fields=["name", "numbers"]`
kz_view fileds option: https://github.com/2600hz/kazoo-core/pull/79
some fields fixes: https://github.com/2600hz/kazoo-core/pull/88


src/crossbar_view.erl
```diff
@@ -370,31 +363,142 @@ has_qs_filter(Context, Options, _, 'false') ->
...
-        'true' -> props:set_value('lager_prefix', "(chunked)", Options);
-        'false' -> Options
+        'true' -> {Context, props:set_value('lager_prefix', "(chunked)", Options)};
+        'false' -> {Context, Options}
+    end.
+
+add_filtermap_fields_options({Context, CursorOptions0}, CbViewOptions) ->
+    FieldKey = props:get_value('field_key', CbViewOptions, <<"doc">>),
+    Fields = get_fields(Context),
+    HasFields = Fields =/= []
```
jamesaimonetti: What happens when `Fields` is `undefined` ? Should `HasFields` be `true`?

src/crossbar_view.erl
```diff
@@ -370,31 +363,142 @@ has_qs_filter(Context, Options, _, 'false') ->
...
+              ],
+    %% in case usermap is a fun/3, store some variables
+    NewContext = cb_context:setters(Context, Setters),
+    FilterMap = crossbar_filter:build_with_mapper(NewContext, UserMapper, HasQSFilter),
+
+    CursorOptions = [{'filtermap', FilterMap}
+                    | props:set_value(['user_params', 'has_fields'], HasFields, CursorOptions0)
+                    ],
+    CursorFields  = [{'field_key', FieldKey}
+                    ,{'fields', Fields}
```
jamesaimonetti: `fields()` type in kz_view does not allow `undefined` which `get_fields(Context)` above allows to be returned.

I think you should call a function after `get_fields(Context)` and branch based whether `get_fields/1` returned `undefined` vs `[]` perhaps?

src/crossbar_view.erl
```diff
@@ -370,31 +363,142 @@ has_qs_filter(Context, Options, _, 'false') ->
...
+    case [normalize_field(F) || F <- List,
+                                is_valid_field(F)
+         ]
+    of
+        [] -> 'undefined';
+        Fields ->
+            %% technically we don't need to _id here, kz_view will it for us,
+            %% but we are doing this, because we store fields in context
+            %% so modules like cb_cdrs can fetch it when generating CSV headers.
+            case props:get_value(<<"_id">>, Fields) of
```
jamesaimonetti: Why not use `lists:member/2` here? Fields isn't a proplist afaict, but `[binary() | [binary()]]` ?

src/crossbar_view.erl
icehess: updated

src/crossbar_view.erl
icehess: Leftover from when Fields could be possibly a prop in crossbar_view.

jamesaimonetti:
cd_cdrs appears to fail when `maybe_filter_empties` returns JSON objects. `map_rows/3` returns an object in that case which `normalize_cdr/3` tries to convert from a list.

jamesaimonetti:
@icehess there are quite a few complaints in crossbar_view if you dialyze-hard:

```
/applications/crossbar/src/crossbar_view.erl:213: warn_matching
  crossbar_view.erl:213: The pattern
          {'ok', {Context, CursorOptions}} can never match the type
          {'error',
           #cb_context{content_types_provided :: [any()],   ...}}
```
for instance.

I suspect the CursorOptions is the culprit. Are you sure it should be `kz_view:options()`? Or should it be the local `options()` type?

Also, can you create some tests in a properly module to show the fieldset work being used?

jamesaimonetti:
`key_min_length` usage suggests using local `options()`

icehess:
> @icehess there are quite a few complaints in crossbar_view if you dialyze-hard:
> 
> ```
> /applications/crossbar/src/crossbar_view.erl:213: warn_matching
>   crossbar_view.erl:213: The pattern
>           {'ok', {Context, CursorOptions}} can never match the type
>           {'error',
>            #cb_context{content_types_provided :: [any()],   ...}}
> ```
> 
> for instance.
> 
> I suspect the CursorOptions is the culprit. Are you sure it should be `kz_view:options()`? Or should it be the local `options()` type?
> 
> Also, can you create some tests in a properly module to show the fieldset work being used?

No it is `kz_view:options()`. The `kz_either:state/3` spec is wrong.

https://github.com/2600hz/kazoo-core/pull/95

icehess:
> `key_min_length` usage suggests using local `options()`

hmm `key_min_length` would be pass to `kz_couch_view` and will be saved in `other_options` in cursor. Also look at the `kz_view:64`

PR #53, opened 2020-09-22
PISTON-1100: Adds voicemail play_envelope_action configuration option.

tetrachromium:
Adds a `play_envelope_action` option to the `system_config.voicemail`, `account_config.voicemail` and `vmboxes` schemas. This is an enumeration of three possible values: `play_after` (the original behaviour), `play_before` (requested by users), and `skip`.

A `skip_envelope` option was already implemented, but represented by a `boolean` value in the schemas. The `skip_envelope` option is mutually exclusive with the new `play_envelope_action` option and isn’t inherited. It has been kept and it takes precedence over the `play_envelope_action` when it is `true`. Otherwise, the behaviour can now be inherited through the new option.

These pull requests are interdependent:
https://github.com/2600hz/kazoo-callflow/pull/17
https://github.com/2600hz/kazoo-crossbar/pull/53
https://github.com/2600hz/kazoo-core/pull/83


jamesaimonetti:
PRs only for master please.

PR #54, opened 2020-09-23
PISTON-1103: Adds voicemail after_forward_action configuration option.

tetrachromium:
An `after_forward_action` configuration option has been added to the `system_config.voicemail`, `account_config.voicemail` and `vmboxes` schemas. The available options are:

* `prompt`: Replays the message menu and asks the user what to do (requested by users).
* `save`: Immediately saves the message after it is forwarded and goes on to the next message (original behaviour).

Under normal circumstances, the `prompt` or `save` option from configuration determines what happens. There are, however, various interactions between forwarding options and edge cases that override configuration. In particular, the user can cancel forwarding, in which case they are always taken back to the message menu. On a hangup or other error, a message is always automatically saved (matches the original implementation).

These pull requests are interrelated and will need to be merged at the same time:
https://github.com/2600hz/kazoo-core/pull/86
https://github.com/2600hz/kazoo-crossbar/pull/54
https://github.com/2600hz/kazoo-callflow/pull/18


jamesaimonetti:
PRs only for master please

PR #58, opened 2020-10-06
PROD-330: patch to documentation of `fields` parameter 

nicholasareed:
Adding additional clarification on usage of `fields` querystring parameter


icehess:
Please update the PR itself https://github.com/2600hz/kazoo-crossbar/pull/50

PR #63, opened 2020-10-14
[master] KCRO-29: return statuses of all devices

icehess:
The device status API currently only returns registered devices. This change this behavior
by returning all devices, indicating registered or not when appropriate.

Additionally, some device types will never be expected to be registered. This list of device types are
stored in System Config `always_registered_device_types` in `crossbar.devices`.
By default this includes:

* application
* cellphone
* sip_url
* landline
* smartphone

Additionally, `/devices` is also changed to accept `?with_status=true` query
string so the user can request registered statues alongside listing devices to
save round trip of listing devices and requesting statues.

### Updates

* add a new fields `registrable` to indicate if device is able to register or not
* Use `registrations.search_req` for query Kamailio for list of registered devices

### Required PRs
Updating proper test: https://github.com/2600hz/kazoo-properly/pull/23
Adding kapi module for `search_req`: https://github.com/2600hz/kazoo-core/pull/110
Updating view to return sip username from: https://github.com/2600hz/kazoo-core/pull/115


src/modules/cb_devices.erl
```diff
@@ -325,10 +342,41 @@ load_device(DeviceId, Context) ->
...
+              ,{'mapper', fun(J, A) -> normalize_reg_devices(J, A, RegIds, RegTypes, ValueFun) end}
+              ],
+    crossbar_view:find(Context, View, Selector, Options).
+
+-spec normalize_reg_devices(kz_json:object(), kz_json:objects(), gb_sets:set(), kz_term:ne_binaries(), fun((kz_json:object()) -> kz_json:object())) ->
+          kz_json:objects().
+normalize_reg_devices(JObj, Acc, RegIds, RegTypes, ValueFun) ->
+    Id = kz_doc:id(JObj),
+    Resp = ValueFun(JObj),
+    case {gb_sets:is_member(Id, RegIds)
```
jamesaimonetti: Isn't this just `gb_sets:is_member(Id, RegIds) orelse lists:member(kz_json:get_value([...], JObj), RegTypes)` ?

src/modules/cb_devices.erl
```diff
@@ -339,6 +387,10 @@ load_device_status(Context) ->
 %%------------------------------------------------------------------------------
 -spec lookup_regs(kz_term:ne_binary()) -> kz_json:objects().
 lookup_regs(AccountRealm) ->
+    lookup_regs(AccountRealm, 'list').
+
+-spec lookup_regs(kz_term:ne_binary(), 'sets' | 'list') -> kz_json:objects().
```
jamesaimonetti: Can return `gb_sets:sets()` too

src/modules/cb_devices.erl
```diff
@@ -352,25 +404,26 @@ lookup_regs(AccountRealm) ->
             lager:debug("error getting reg: ~p", [_E]),
             [];
         {_, JObjs} ->
-            [kz_json:from_list([{<<"device_id">>, AuthorizingId}
-                               ,{<<"registered">>, 'true'}
-                               ])
-             || AuthorizingId <- extract_device_registrations(JObjs)
-            ]
+            case RetType of
```
jamesaimonetti: What about

```erlang
            process_regs(JObjs, RetType)
     end.
 
process_regs(JObjs, 'sets') ->
    extract_device_registrations(JObjs, gb_sets:new());
process_regs(JObjs, 'list') ->
    [kz_json:from_list([{<<"device_id">>, AuthorizingId}
                       ,{<<"registered">>, 'true'}
                       ])
     || AuthorizingId <- gb_sets:to_list(process_regs(JObjs, 'sets'))
    ].

```

src/modules/cb_devices.erl
```diff
@@ -256,20 +268,25 @@ load_device_summary(Context) ->
...
                ],
-    crossbar_view:find(Context, <<"crossbar_listings/by_type_id">>, Selector, Options).
+    WithRegs = kz_term:is_true(cb_context:req_param(Context, <<"with_status">>)),
+    maybe_with_device_status(Context, <<"crossbar_listings/by_type_id">>, Selector, WithRegs).
+
+maybe_with_device_status(Context, View, Selector, 'true') ->
+    ValueFun = fun(JObj) -> kz_json:get_value(<<"value">>, JObj) end,
+    load_device_status(Context, View, Selector, ValueFun);
+maybe_with_device_status(Context, View, Selector, 'false') ->
+    Options = [{'doc_type', <<"device">>}
```
jamesaimonetti: Can we use `kzd_devices:type()` in these places where the doc type is used?

doc/devices.md
```diff
@@ -278,11 +280,48 @@ curl -v -X GET \
...
+            "features": [],
+            "owner_id": "96f99542f33c180790d1836192300c27",
             "device_type": "sip_device",
-            "enabled": false,
-            "id": "{DEVICE_ID}",
-            "mac_address": "00:04:f2:ab:7e:fd",
-            "name": "MyPolycom"
+            "mac_address": "001122334455",
+            "username": "user_meme",
+            "registered": true
```
jamesaimonetti, line 326: Personally I'm not crazy on the idea of `registered: true` for non-registering devices. Seems like we have a tristate situation: `is_registered`, `not_registered`, `doesnt_register`. Not sure the best format to represent this enum for JSON. Thoughts?

jamesaimonetti (changes requested):

doc/devices.md
joristirado: How about the absence of the `registered` prop to denote devices that can't register?

jamesaimonetti: Perhaps two keys?:
1. `registered: true, registerable: true` - when registered
2. `registered: false, registerable: true` - when not registered
3. `registerable: false` - when not applicable

Or LISPy `registersp` :)

jamesaimonetti:
Added test https://github.com/2600hz/kazoo-properly/pull/29 for non-registering devices too

PR #69, opened 2020-10-30
PISTON-1143: Reconciles the voicemail schema documentation.

tetrachromium:
At some point, the `voicemail` document was renamed as `vmboxes`, but the related documentation files were not  renamed to match. As a result, a hand-edited `voicemail.md` document diverged considerably from the auto-generated `ref/vmboxes.md` document. The contents of the `voicemail.md` have been reordered to reduce the degree of difference from the auto-generated file, and some minor corrections were made to the text.

Kazoo 4.3 pull request: https://github.com/2600hz/kazoo/pull/6656


tetrachromium:
Looking at the differences within the `voicemail.md` file is likely to just be confusing. To understand the reasons for the changes and check for correctness, the revised `voicemail.md` file should be compared with the `ref/vmboxes.md` file. I suspect that the amount of difference between the two files has crept in due to the fact that they no longer have the same name. It was a while before I realized the two files were related. If you are willing to rename the file and redirect from the old URL, let me know and I will do that as a separate commit.

doc/voicemail.md
```diff
@@ -372,80 +328,97 @@ curl -v -X PUT \
...
 
-## Fetch all messages for a voicemail box
+## Create a new voicemail message
 
-> GET /v2/accounts/{ACCOUNT_ID}/vmboxes/{VM_BOX_ID}/messages
+> PUT /v2/accounts/{ACCOUNT_ID}/vmboxes/{VM_BOX_ID}/messages
+
+There are two methods for creating a new voicemail message - they differ in how you attach the media file.
+
+In the first method, you can create a voicemail document first in one request and then put the media file into the document with a second request using `/messages/{VM_MSG_ID}` API endpoint.
```
icehess: It is actually `PUT /messages/{VM_MSG_ID}/raw`.

icehess: make a link to the section:

```
[`/messages/{VM_MSG_ID}/raw`](create-the-raw-audio-for-a-voicemail-message)
```

doc/voicemail.md
```diff
@@ -464,27 +436,30 @@ curl -v -X GET \
...
+            "media_id": "201605-fadnew0mf6fcfgfd8bcdfca312e924bq"
+    },
     "revision": "{REVISION}",
     "request_id": "{REQUEST_ID}",
     "status": "success"
 }
 ```
 
-## Change a list of messages
+And then you can use PUT method on `/messages/201605-fadnew0mf6fcfgfd8bcdfca312e924bq` to add the media to file (see PUT method for a message below).
```
icehess: same as above it has `/raw` at the end

```
[`/messages/{VM_MSG_ID}/raw`](create-the-raw-audio-for-a-voicemail-message)
```

doc/voicemail.md
```diff
@@ -372,80 +328,97 @@ curl -v -X PUT \
...
-Deleting all message is easy, just use `DELETE` method on message API endpoint to delete all account's messages.
-
-Optional payload for deleting a group of messages:
-
-* One can apply a filter to delete all messages in a particular folder(e.g. new or saved) by adding a query string `?folder=saved` to the URL or set it in the payload as `{"data": {"folder": "saved"}}`
-* Or providing an array of message ids, e.g `{"data": {"messages": [MSG_ID1, MSG_ID2, ...]}}`.
-
-!!! note
-    If you didn't move voicemail messages to the new format already, messages that are in old format will be moved to the new MODB format, which will cause their message id to change to the new format.
+With pagination:
```
icehess, line 361: It should do pagination by default. Did you mean **without** pagination?

icehess (changes requested):

doc/voicemail.md
tetrachromium: The query parameter is `pagination=true` so "without pagination" wouldn't seem to make much sense. My goal is simply to get this file to more or less match the `ref` version of the file again. If there are documentation errors, they were there already. You will have to tell me what corrections you want or can make them yourselves after this is merged.

doc/voicemail.md
tetrachromium: Changed.

doc/voicemail.md
tetrachromium: Changed.

icehess (approved):

PR #70, opened 2020-11-04
files generated from "make apis", need it to complete CI on PR "Kftp …

oguerrero913:
…24 system config #20" for kazoo-cdr-push repo


oguerrero913:
added {schemas_to_priv, true} in the env props in app.src file

PR #74, opened 2020-11-12
KCRO-45: fixed api reply

holy-batman:
fixed number delete api results


src/modules/cb_phone_numbers.erl
```diff
@@ -938,9 +938,9 @@ set_response({'dry_run', Quotes}, Context, CB, _Collection) ->
     end.
 
 -spec set_error(knm_errors:reason(), cb_context:context()) -> cb_context:context().
```
icehess: change `knm_errors:reason()` to `{kz_term:ne_binary(), knm_errors:reason()}`

icehess (changes requested):

src/modules/cb_phone_numbers.erl
holy-batman: sorry for the rookie mistake, changed it.

icehess: hehe, no worries. Good cacth btw how this thing was working before?? :)

src/modules/cb_phone_numbers.erl
```diff
@@ -937,10 +937,10 @@ set_response({'dry_run', Quotes}, Context, CB, _Collection) ->
...
--spec set_error(knm_errors:reason(), cb_context:context()) -> cb_context:context().
-set_error('not_found', Context) ->
+-spec set_error({kz_term:ne_binary(), knm_errors:reason()}, cb_context:context()) -> cb_context:context().
+set_error({_, 'not_found'}, Context) ->
     reply_number_not_found(Context);
-set_error(Data, Context) when is_atom(Data) ->
+set_error({_, Data}, Context) when is_atom(Data) ->
     lager:debug("error: ~p", [Data]),
     crossbar_util:response_400(<<"client error">>, Data, Context);
 set_error(Data, Context) ->
```
icehess: do we need to change this too? base on knm_numbers:return()? also check `927: set_response` line too

src/modules/cb_phone_numbers.erl
holy-batman: it was using the next fn and going to error 500 as it was receiving unexpected data

src/modules/cb_phone_numbers.erl
holy-batman: I thought, this is getting json data but that's why was confused with spec, lemme do some more tests to check data. so according to spec this will also be changed. thank you.

src/modules/cb_phone_numbers.erl
harenson: I think it should be `knm_errors:reason() | {kz_term:ne_binary(), knm_errors:reason()}`

holy-batman:
@icehess the change was minor, checked the Fns as well as the fn you have mentioned.

holy-batman:
this essentially fixes, 
KCRO-42
KCRO-43
KCRO-44
KCRO-45

Testing all of these to be 100% sure will move them to ready for QA after testing and when this is merged

PR #75, opened 2020-11-15
KCRO fixes

icehess:
This fixes some kazoo view cursor and crossbar view issues and others.

* Let parse `fields` query string properly and allow passing JSON as value, and normalize the keys and values:
    * return error if failed to parse or have erroneous value
    * Allow `_read_only` in fields
    * We're still just ignore other private fields


src/crossbar_view.erl
```diff
@@ -513,54 +516,109 @@ module_allows_include_docs([{Endpoint, _} | _]) ->
...
-    Fields = [normalize_field(F) || F <- List,
-                                    is_valid_field(F)
-             ],
     %% technically we don't need to add _id here, kz_view will add it for us,
-    %% but we are doing this because we store fields in context modules like
-    %% cb_cdrs can use normalized fields from context when generating CSV headers.
-    case lists:member(<<"_id">>, Fields) of
+    %% but we are doing this because we store fields in context and to let modules like
+    %% cb_cdrs use normalized fields from context when  need it.
+    case normalize_fields(List, []) of
```
jamesaimonetti: Is there any danger to just including `<<"_id">>` in List here and `lists:usort(Fields)` without the membership check?

If `{<<"_id">>, bool()}` is included, you already double up so it seems like you can just force `<<"_id">>` to be present.

src/crossbar_view.erl
```diff
@@ -513,54 +516,109 @@ module_allows_include_docs([{Endpoint, _} | _]) ->
...
+            case lists:member(<<"_id">>, Fields) of
+                'false' ->
+                    lists:usort([<<"_id">> | Fields]);
+                'true' ->
+                    lists:usort(Fields)
+            end
+    end.
+
+-spec normalize_fields(kz_view:fields(), kz_view:fields()) -> kz_view:fields().
+normalize_fields([], Acc) ->
```
jamesaimonetti, line 570: Do you prefer this manual definition of a folding function? Personally, I think `lists:foldl(fun normalize_field/2, [], List)` and defining this function as `normalize_field(<<"_id">>, Acc) -> [<<"_id">> | Acc]);...` keeps the functions focused on normalization of a field vs normalizing plus recursion.

I don't know that it makes a difference to the runtime but just curious why you implement the folding too.

jamesaimonetti (changes requested):

src/crossbar_view.erl
icehess: hmm, I should use `props:get_value(<<"_id">>, List)` as same as `kz_view`.

src/crossbar_view.erl
```diff
@@ -510,54 +513,109 @@ module_allows_include_docs([{Endpoint, _} | _]) ->
...
-    Fields = [normalize_field(F) || F <- List,
-                                    is_valid_field(F)
-             ],
     %% technically we don't need to add _id here, kz_view will add it for us,
-    %% but we are doing this because we store fields in context modules like
-    %% cb_cdrs can use normalized fields from context when generating CSV headers.
-    case lists:member(<<"_id">>, Fields) of
+    %% but we are doing this because we store fields in context and to let modules like
+    %% cb_cdrs use normalized fields from context when  need it.
+    case lists:foldl(fun normalize_fields/2, [], List) of
```
jamesaimonetti, line 558: I guess my previous question was, can this be turned into:

```erlang
case lists:foldl(fun normalize_fields/2, [], [<<"_id">> | List]) of
    [] -> 'undefined';
    Fields -> lists:usort(Fields)
end
```

icehess: No, because I want to allow client to request to not add `id` with `{<<"id">>, false}`. Similar to MongoDB which by default is adding `id` but can turn if off.

PR #78, opened 2020-11-24
KCRO-29: fix status, realm and username must be lower case

icehess:
Fixing an issue when listing device registration statuses. When sending registration search request and checking its response, usernames and account realm must be lower case.


lazedo:
@icehess username & realm don't need to be `lowercase` and we shouldn't do this.

icehess:
Per discussion:

* Usernames are case-sensitive according to RFC [1]
* Account realm MUST be lower case according to RFC [1] (other parts after username of SIP URI is case-insensitive)
* Right now we don't set `case_sensitive` params in Kamailio however if that parameter is changed in the future we better to lower case the username (and realm for accounts that has upper case realms) to be safe here

[1] https://tools.ietf.org/html/rfc3261#section-19.1.4

PR #81, opened 2020-11-26
[KCRO-62]: fix unassigning number from callflow won't update the number docs

icehess:
With update to how we validate callflow, old old callflow won't be store in
context and track_assignment function wont detect any removal when calling knm
modules.


holy-batman:
The fix is working as excepted, can be merged. CI being bad again.

PR #85, opened 2020-11-30
KZOO-87: if decoding of start_key fails, return empty list

jamesaimonetti:
If `start_key="undefined"` was in the query string, `bookmark` would
get set to <<"undefined">>. When crossbar_view tried to get the
`next_startkey` from `bookmark`, props would fail as the `bookmark`
key is not a list:

`props:get_value(['bookmark','next_startkey'], Options)` when
`{'bookmark', <<"undefined>>}`.


jamesaimonetti:
See test in https://github.com/2600hz/kazoo-properly/pull/35

Before this change, the test would return a 500 error after crashing on the props issue. Now returns 404 as expected.

icehess (approved):

PR #96, opened 2020-12-10
Fetch provisioner data using typed accessors

jamesaimonetti:
When validating creating/updating a new device, `validate_device/2`
has a fold over functions that "fix up" request data before doing
schema validation.

When `prepare_provisioner_fields/2` runs, if the request data's
`provisioner` object has invalid data for `combo_keys` or
`feature_keys`, the code will crash in
`prune_null_provisioner_fields/2`. In the observed case, `combo_keys`
was a list of objects instead of the expected JSON object. The API
returns an init_failed error instead of a schema validation error.

The change is to specify the kzd_devices getter and setter functions
for those fields. The getter will ensure the data is of the correct
shape or else return `undefined`, leaving the bad data untouched.

The `validate_device/2` fold can then reach schema validation and
correctly return the validation error.

See https://github.com/2600hz/kazoo-properly/pull/39 for the test
of an expected validation error.


jamesaimonetti:
The code tried to fetch the `provision` object using `cb_context:req_value/2` which searched the request data, querystring, and request envelope, in order. Because of this, the typed accessors of kzd_devices were of no use. I don't see any examples of including a `provision` JSON object on the querystring and certainly it doesn't make sense to be on the envelope itself.

By only concerning ourselves with the request data object, we can use the accessors to help get correctly-shaped data.

Is this a correct assumption?

jamesaimonetti:
The other question is can the folded functions in `validate_device/2` be re-arranged to run schema validation first?

k-anderson:
@jamesaimonetti lgtm, regarding looking for provisioning data in the envelope or query string, agreed it would never be expected there.  With regards to validating the schema first, I am of the opinion that its better to validate the document after manipulation so you are validating what will actually be stored... however, if there is a reason that you would prefer the otherway round let me know :)

PR #112, opened 2020-12-20
FORUM-11759: only set id from req data if allowed by system

jamesaimonetti:
Way back in the day (31 Oct 2012!), `onboarding` was allowed to set
`_id` on the cb_context.doc field from the `_id` on the request data if
included.

Six years ago (7 July 2015) a change to use accessors (wh_doc:id/1 in
this case) was made. Two months prior (15 May 2015), `wh_doc:id/1` was
updated to check for `_id` then `id`. Now if the request data included
either `_id` or `id`, it would be included on the resulting
cb_context.doc and used as `_id` in the database.

It seems like this isn't in the spirit of the pvt fields and, as far
as I know, the onboarding app is not a thing.

Now why the 'users' API does not honor this while other APIs do, I'm
not sure yet but it felt like maybe the broader conversation about
whether this should continue to be allowed should be had first!


jamesaimonetti:
See https://github.com/2600hz/kazoo-properly/pull/42

joristirado:
From my standpoint, I don't see any problem with either behavior, assuming the custom `id` prop setting only takes effect on document creation. If it also affect document updates, there might some places in the UI where we send the full document with an `id` prop that MIGHT NOT correspond to the original one (highly unlikely though).

Also not sure why someone would provide a custom `id` in the first place, the only reason I could think of is performance.

PR #115, opened 2020-12-30
KCRO-85: Fix crash when transaction id is invalid.

holy-batman:
we have deprecated some transaction id functions but seems like customers are still using those, this will make sure we give them proper error code instead of just crashing.


src/modules/cb_transactions.erl
```diff
@@ -170,14 +170,17 @@ validate_transaction(Context, ?REFUND, ?HTTP_PUT) ->
...
     case kz_transaction:fetch(AccountId, Id) of
         {'ok', Transaction} ->
             crossbar_doc:handle_json_success(kz_transaction:public_json(Transaction), Context);
         {'error', Reason} ->
             crossbar_doc:handle_datamgr_errors(Reason, Id, Context)
-    end.
+    end;
+validate_transaction(Context, _, ?HTTP_GET) ->
+    Message = kz_json:from_list([{<<"message">>, <<"invalid transaction id">>}]),
+    cb_context:add_validation_error(<<"transaction_id">>, <<"invalid">>, Message, Context).
```
icehess: Probably it is better to use the standard bad identifier error:

```suggestion
validate_transaction(Context, Id, ?HTTP_GET) ->
    crossbar_util:response_bad_identifier(Id, Context).
```

icehess (changes requested):

holy-batman:
@icehess done, thank you for the review!

PR #117, opened 2021-01-05
KCAL-22: Add new parameter to exclude years from temporal_rules

harenson:
Related:
https://github.com/2600hz/kazoo-core/pull/188 (Required for CI checks)
https://github.com/2600hz/kazoo-callflow/pull/49

Before this change if you wanted to have a recurring holiday that should happen one day every year but not on a given year you would need to create 2 rules, one to start (maybe) now until before the unwanted year and another to start after the unwanted year until the end of the days.

Let us use the *Labor Day* (first Mnday of September)  as an example. Suppose we want Labor day to happen every year but on September 2022; before this change you would need to create 2 rules, one to be a single day rule to happen on *20210906* and another one to be a recurring holiday that would start on *20230904* until infinity. After this change you can create one single rule and just set the `exclude` key to contain *20220905* that way that rule will be applied every year but on September 2022.

Look at https://github.com/2600hz/kazoo-crossbar/blob/master/doc/temporal_rules.md#exclude for an example.


src/api_util.erl
```diff
@@ -200,7 +200,7 @@ get_query_string_data([], Req) ->
     {kz_json:new(), Req};
 get_query_string_data(QS0, Req) ->
     QS = kz_json:from_list(QS0),
-    lager:debug("query string: ~s", [kz_json:encode(QS, ['pretty'])]),
+    lager:debug("query string: ~s", [kz_json:encode(QS)]),
```
icehess: Why this change? revert it.

icehess (changes requested):
Revert the pretty print QS.

Also add an example date format for exclude so people know how to use it.

icehess (approved):

PR #120, opened 2021-01-14
KCDR-1: Filter pvt fields in ccvs

holy-batman:
this is a temp fix to make sure we are not getting pvt fields in CCVS


src/modules/cb_cdrs.erl
```diff
@@ -579,3 +581,30 @@ normalize_accept_type([_Accept|Accepts]) ->
...
+    cb_context:set_resp_data(Context, remove_pvt_ccvs([JObj], [])).
+
+-spec filter_pvt_fields_legs(cb_context:context()) -> cb_context:context().
+filter_pvt_fields_legs(Context) ->
+    JObjs = cb_context:resp_data(Context),
+    cb_context:set_resp_data(Context, remove_pvt_ccvs(JObjs, [])).
+
+-spec remove_pvt_ccvs(kz_json:objects(), kz_json:objects()) ->
+          kz_json:objects().
+remove_pvt_ccvs([JObj| JObjs], Acc) ->
```
jamesaimonetti: I would just use list comprehension to iterate over the list instead of the accumulator:

```erlang
filter_pvt_fields(JObjs) ->
    [remove_pvt_ccvs(JObj) || JObj <- JObjs].

remove_pvt_ccvs(JObj) ->
    CCVs = kz_json:get(),
    FilteredCCVs = kz_json:foldl(),
    kz_json:set_value(ccvs, FilteredCCVs, JObj).
```

src/modules/cb_cdrs.erl
```diff
@@ -579,3 +581,30 @@ normalize_accept_type([_Accept|Accepts]) ->
...
+    CCVs = kz_json:get_json_value(<<"custom_channel_vars">>, JObj, kz_json:new()),
+    CCVs1 = kz_json:foldl(fun update_ccvs_foldl/3, CCVs, CCVs),
```
jamesaimonetti: You might also use `kz_doc:public_fields(CCVs)` instead of rolling your own?

src/modules/cb_cdrs.erl
```diff
@@ -535,7 +536,8 @@ load_legs(<<BinTimestamp:11/binary, "-", _Key/binary>>=InteractionId, Context) -
               ,{'max_range', ?MOD_MAX_RANGE}
               ,'include_docs'
               ],
-    crossbar_view:load_modb(Context, ?CB_INTERACTION_LIST_BY_ID, Options);
+    Context1 = crossbar_view:load_modb(Context, ?CB_INTERACTION_LIST_BY_ID, Options),
+    filter_pvt_fields_legs(Context1);
```
icehess: As general rule, if it possible NEVER call any function after `crossbar_view` which is manipulating resp in any way. Unless you explicitly turn of chunking/batching. In this case I won't think anyone gonna use chunking, but it is a good practice to avoid these situation.

One way to to do what you want is to do it in the `mapper` function. Create another mapper function which it turn calls `filter_pvt_fields_legs`.

icehess (changes requested):

src/modules/cb_cdrs.erl
holy-batman: @icehess thank you for the tip. is there any reference or example code I can look at which can give an idea on how it works with mapper function.

icehess: No doc, sorry. But just grep for `mapper` in crossbar, there is plenty usage there.

Basically the function can be these:
* `fun normalizer/1`: which crossbar_view calls it with all JObj, `normalizer(JObjs)`
* `fun normalizer/2`: typical Erlang list tail recursion `normalizer(JObj, Acc)`
* `fun normalizer/3`: `crossbar_view` calls it like `/2` above but Context: `normalizer(Context, JObj, Acc)`

There are some other stuff to know, like don't change the order of `Acc or JObjs`, `mapper` is always applies after corssbar_filter (look at crossbar_filter:build_filter_map_fun/3` to see how the it is construct.

If you just want to get typical Couch view result fields, crossbar_view has handy function for mapper:
`[{'mapper', crossbar_view:get_doc_fun()}]`, for doc. Use `get_value_fun` and etc.. for the rest.

icehess:
We do remove private fields from ccvs, in both `cdr_channel_destroy` and `cdr_report`. Thinking why this is needed after all?

src/modules/cb_cdrs.erl
icehess: `crossbar_view` with the help of `crossbar_filter` will eventually translate the `mapper` option to the `filtermap` option of `kz_view`.

filtermap is almost same as mapper. To see what you can return in your mapper function take a look at it `kz_view` file.

holy-batman:
> We do remove private fields from ccvs, in both `cdr_channel_destroy` and `cdr_report`. Thinking why this is needed after all?

@icehess Yes we do and that code seems to be working perfectly fine, but somehow since 4.3 we are still seeing pvt field in cdr ccv, this will take care of any existing cdrs and won't show pvt values for those.

src/modules/cb_cdrs.erl
holy-batman: @icehess fixed

icehess (approved):

PR #123, opened 2021-01-15
KZOO-120: Add carrier ACL when a local carrier is added.

holy-batman:


src/modules/cb_resources.erl
```diff
@@ -300,15 +294,14 @@ validate_jobs(Context, ?HTTP_PUT) ->
 
 -spec post(cb_context:context(), path_token()) -> cb_context:context().
 post(Context, ?COLLECTION) ->
-    maybe_reload_acls(cb_context:db_name(Context)),
+    _ = reload_trusted(),
```
lazedo: reload should be done on success not before

src/modules/cb_resources.erl
```diff
@@ -352,26 +342,24 @@ delete(Context, ResourceId) ->
 %% @end
 %%------------------------------------------------------------------------------
 -spec maybe_aggregate_resource(cb_context:context(), kz_term:ne_binary(), crossbar_status()) -> cb_context:context().
-maybe_aggregate_resource(Context, ?KZ_OFFNET_DB, _) -> Context;
+maybe_aggregate_resource(Context, ?KZ_OFFNET_DB, _) ->
+    _ = reload_trusted(),
```
lazedo, line 345: reload should be done `once` on successful completion

src/modules/cb_resources.erl
```diff
@@ -318,13 +311,11 @@ patch(Context, Id) -> do_post(Context, Id).
 -spec put(cb_context:context()) -> cb_context:context().
 put(Context) ->
     Db = cb_context:db_name(Context),
-    maybe_reload_acls(Db),
     Context1 = crossbar_doc:save(Context),
     maybe_aggregate_resource(Context1, Db, cb_context:resp_status(Context1)).
 
 -spec put(cb_context:context(), path_token()) -> cb_context:context().
 put(Context, ?COLLECTION) ->
-    maybe_reload_acls(cb_context:db_name(Context)),
```
lazedo, line 327: processing a collection the way it is now would `reload` the number of the docs

lazedo (changes requested):
reload should be done on successful completion, not on each doc (in case of collection)

src/modules/cb_resources.erl
holy-batman: changed to only refresh if success.

src/modules/cb_resources.erl
holy-batman: changed collection refresh to be done after a success is received from collection.

lazedo (approved):

src/modules/cb_resources.erl
```diff
@@ -652,119 +516,21 @@ aggregate_resource(ResourceJObj) ->
...
-%%------------------------------------------------------------------------------
-%% @doc
-%% @end
-%%------------------------------------------------------------------------------
--spec maybe_remove_aggregates(kzd_resources:docs()) -> 'ok'.
-maybe_remove_aggregates([]) -> 'ok';
-maybe_remove_aggregates([ResourceJObj|ResourceJObjs]) ->
-    _ = remove_aggregate(kz_doc:id(ResourceJObj)),
-    maybe_remove_aggregates(ResourceJObjs).
+    _ = aggregate_resource([ResourceJObj]),
```
jamesaimonetti: Why wrap this in a list? Shouldn't `aggregate_resource/1` take a JSON object only, not a list of JSON objects?

And I'd rather just call `lists:foreach/2` here:

```erlang
maybe_aggregate_resources(ResourceJObjs) ->
    lists:foreach(fun aggregate_resource/1, ResourceJObjs).
```

jamesaimonetti (changes requested):

src/modules/cb_resources.erl
holy-batman: @jamesaimonetti fixed :)

PR #127, opened 2021-01-19
[master] KCRO-89: fix crash when prepare_docs returns error

icehess:


icehess:
5.0 PR: https://github.com/2600hz/kazoo-crossbar/pull/128

PR #128, opened 2021-01-19
[5.0] KCRO-89: fix crash when prepare_docs returns error

icehess:


icehess:
Master PR: https://github.com/2600hz/kazoo-crossbar/pull/127

PR #134, opened 2021-01-29
[5.0] KCRO-91: Handle file not found when deleting file

harenson:
Master: https://github.com/2600hz/kazoo-crossbar/pull/133
Without this change api_util will crash every time it tries to create a file that does not exist.


src/api_util.erl
```diff
@@ -1285,8 +1285,12 @@ cleanup_file(ConnPid, File) ->
 
     wait_for_exit(ConnPid, Ref),
 
-    'ok' = file:delete(File),
-    lager:debug("deleted file ~s after ~pms", [File, kz_time:elapsed_ms(Start)]).
+    case file:delete(File) of
+        'ok' ->
+            lager:debug("deleted file ~s after ~pms", [File, kz_time:elapsed_ms(Start)]);
+        {'error', 'enoent'} ->
+            lager:debug("file ~s not found after ~pms", [File, kz_time:elapsed_ms(Start)])
```
holy-batman: wouldn't this be error or info to show if there could be a potential issue ?

PR #137, opened 2021-02-03
KCRO-92: Add schema validation for timezone

holy-batman:
Add Schema validation for timezone to make sure we are not letting in unsupported Timezone, the list of timezones is retrieved from monster-ui's supported timezone list


holy-batman:
requires https://github.com/2600hz/kazoo-core/pull/207 for ci checks

harenson (approved):
:+1:

lazedo:
@holy-batman why is this needed and why is not handled in a callback ? also, how does the added file compare to the library/dep we're using ?

holy-batman:
> @holy-batman why is this needed and why is not handled in a callback ? also, how does the added file compare to the library/dep we're using ?

Right now we do not validate if the timezone is valid or not due to which we got a ticket few days ago which caused crashes, the crash was happening with the library we use for timezone, essentially right now we can even save timezone as `Mars/Colony1`. 
Here the list of timezones are the ones we support in monster ui and not in the library as library has many more options, if we use any other timezone which is not in monster-ui then we get invalid timezone selection in monster-ui, this will make sure if we want to add support for other timezones we add those to monster as well to make it consistent.

icehess:
@holy-batman look at `applications/crossbar/priv/port-request-scheduled-schema.escript`, This is generating all timezone using `erlang_localtime`.

Change that script to generate timezone schema and reference that schema everywhere that needs timezone.

jamesaimonetti:
@icehess moved the script to generate timezone.json enum and ref'd it in port_request.to_scheduled.json. Let me know if there's more to do here.

@holy-batman let me know if i missed anything else here? Want to add a properly test to check that invalid timezones are rejected and valid ones are accepted? :)

jamesaimonetti:
See https://github.com/2600hz/kazoo-properly/pull/54 for test

icehess (approved):

PR #141, opened 2021-02-11
PISTON-1085: Do not overwrite a system error message if its set when adding the error to a crossbar context

bradfordben:
Kazoo 4.3 PR https://github.com/2600hz/kazoo/pull/6700


jamesaimonetti:
@bradfordben sorry for the delay, what's the API call that you've seen the overwrite happening on? What's the expected vs actual results of that API call? Would like to get a properly test case showing the issue.

jamesaimonetti:
@bradfordben would like to get a test for this; what API request triggers it?

manushiperera:
Hello @bradfordben,

Any update on this?

bradfordben:
Hi, sorry about the delay on getting back to you. This was so long ago i cant quite remember but i feel its something like if you try and update or create a object through crossbar and it has multiple validation errors it will now return all the errors in one json response rather than only the last error. I dont have a working setup at the moment to just try this quickly unfortunately.

PR #145, opened 2021-02-16
KZOO-158: Add commands to delete and rename+refresh apps metadata

harenson:
Before this change whenever an application's name was changed we needed to delete the application's metadata from BigCouch manually and then `init` the application with the new name.

Also, there was not command for deleting old applications if needed.


icehess:
I think we should avoid deleting the apps only for just renaming it.

App's ids are saved in mutiple places from user documents to service plans and other places. If the app is just renamed then just rename it, DO NOT delete it and re-create it.

Only delete an app if it is really deleted. In that case then you should also add some other maintenance commands to find all traces of apps and remove it from there.

harenson:
@icehess that's exactly why I marked the PR as a draft because while talking to Vlad I thought the app may also be associated to other entities like accounts and I wanted to make sure this PR contains all the necessary code to handle those delete actions as well. And just to be clear the *delete* command is not meant to be used for renaming apps, it was added only for those scenarios where and old app is deprecated and needs to be deleted or something alike.

harenson:
Moved to:
- https://github.com/2600hz/kazoo-core/pull/270
- https://github.com/2600hz/kazoo-crossbar/pull/191

vlad88sv:
Hi, can this `delete command` be retaken? we have ran into this situation with `monster-ui-application-integration-duo`.

Situation: upgrade from 4.3 to 5.1:

1. `monster-ui-application-duo` is removed during upgrade 
2. `monster-ui-application-integration-duo` is the new package
3. the issue is that the old app folder `duo` is removed, and a new folder `integration-duo` is created so we can't use `rename_and_refresh_app`
4. we need a way to "delete" the old app, without the physical app folder so we can install the new app

icehess:
@vlad88sv 

1. the package `monster-ui-application-duo` was install, you ran the `init_apps` and kazoo created the doc id `1234` for the app.
2. the `monster-ui-application-integration-duo` is installed which replaced the old app. You should have ran `rename_and_refresh_app` because the app name was changed to `integration-duo` and delete the `duo` folder from file system then run refresh_apps
3. But you didn;t and instead run refresh_apps, now we have two apps: duo and integration-duo
4. since all accounts are tied to the duo app id, you can simply delete duo. You need to delete the old integration-duo, and revert any account that is used to integration-duo back to duo and then run that rename comand again

PR #148, opened 2021-02-23
get filename from content-disposition header if possible

icehess:
If request is `multipart/form-data` try to get `filename` from `content-disposition` header if query string `filename` is not provided. If non is exists, generate a random file name.

`decode_base64` also has been fixed to honor filename in the request.

This also make sure the file name has extension. otherwise will add one based on the part `content-type` header. Also we now properly set all part's headers in req_files.


icehess:
Closing in favor of https://github.com/2600hz/kazoo-crossbar/pull/181

PR #151, opened 2021-02-26
API doc fix callflow endpoints

filipevarjao:
API docs are not specifying the content type and body request to callflow endpoints


doc/callflows.md
```diff
@@ -113,18 +115,26 @@ curl -v -X GET \
 
 > PATCH /v2/accounts/{ACCOUNT_ID}/callflows/{CALLFLOW_ID}
 
+It is possible to pass a single field or the entire callflow object:
```
jamesaimonetti: I think this line and the one for POST below are unnecessary. It is expected that folks will have read https://docs.2600hz.com/dev/applications/crossbar/doc/basics/#http-verbs Only if an endpoint changes these meanings of the verbs should it be noted.

PR #158, opened 2021-03-11
[master] KZOO-163: use by_type_id to get user s summary

icehess:
use doc id as key when listing user summary.


icehess:
5.0: https://github.com/2600hz/kazoo-crossbar/pull/158
4.3: https://github.com/2600hz/kazoo/pull/6707

Core master pr: https://github.com/2600hz/kazoo-core/pull/232
Location master PR: https://github.com/2600hz/kazoo-locations/pull/6

PR #162, opened 2021-03-17
KCRO-94: Fix timestamp values returned by /cdr API

harenson:
Master: https://github.com/2600hz/kazoo-crossbar/pull/161

For cdrs with version 1 and 2 we needed to subtract the *duration* seconds from the call's *timestamp* because they were using  the channel destroy as the timestamp value but this is not required anymore on 5.x because it is using the channel_created_time FS variable as the timestamp value.


lazedo (changes requested):
can we move the logic to `kzd_cdrs:timestamp` instead ? a `scan` must be done in code base for instances using cdr timestamp either with accessors or thru kz_json.

it may be a bit more complex, but will serve tasks app too.

maybe the logic should be added to kzd_cdr:channel_created_time  and have kzd_cdr:timestamp return kzd_cdr:channel_created_time_gregorian 

then make sure channel_created_time is returned in `interaction` listing

lets get @jamesaimonetti opinion on this

jamesaimonetti:
@lazedo sorry, missed your comment! I agree on moving this to kzd_cdrs; will issue correcting PRs.

jamesaimonetti:
@harenson @lazedo actually, we have `kzd_cdrs:channel_created_time/1` now; why not just use that in cb_cdrs?

PR #163, opened 2021-03-18
PROD-145: rework scope and scope restrictions to deny access base on rules

icehess:
All adding scope to JWT token and its authorization are moved to the code.

This will update scope functionality in a way that all application can use.

When creating JWT, modules all defined scope on request (if allowed), user doc, account doc and available scopes from system will be added to JET token.

During authorization, application can call `kz_auth_scope:maybe_deny_access(AuthDoc, Options)` to check if request is authorized based on scopes defined in the token and their rules.

Scope rules (scope restrictions) are defined for each scope in a document system wide or in account db. The rules may support multiple transport. Currently we have `http`. Other possible transport like `websocket`, `sip` may be added in the future if needed.


icehess:
Required PRs:
* https://github.com/2600hz/kazoo-core/pull/236

manushiperera:
Hello @icehess,
Since this is an old PR, are we still good at merging these changes or are they already depreciated? 
Any thoughts?

icehess:
@manushiperera Please ignore this PR for time being

manushiperera:
Hello @icehess, understood. Thanks for the update

PR #169, opened 2021-03-23
KCRO-95: leak smtp_email_address pvt field for /faxboxes/ID requests

harenson:
Also:

- Leak pvt fields for /faxboxes API in the crossbar way.
- Add single_doc normalization feature to crossbar_doc.
- Fix PATCH request validation.

5.0: https://github.com/2600hz/kazoo-crossbar/pull/170

On 4.3 the `smtp_email_address` private field was being leaked when the details for a given faxbox was requested (`GET /faxboxes/{ID}`) and this behaviour was unintentionally changed on 5.x. This PR puts that behaviour back in place and also adds the functionality on *crossbar_doc* module to handle normalize functions for single documents.


src/crossbar_doc.erl
```diff
@@ -49,8 +49,10 @@
                   ,fun add_pvt_alphanum_name/3
                   ]).
 
+-type normalize_doc_fun() :: fun((kz_json:object()) -> kz_json:object()).
 -type load_option() :: {?OPTION_EXPECTED_TYPE, kz_term:ne_binary()} |
-                       {'use_cache', boolean()}.
+                       {'use_cache', boolean()} |
+                       {'normalize_doc', normalize_doc_fun()}.
```
jamesaimonetti: This property makes me think `normalize_doc` is a `boolean()`. Maybe `doc_normalizer`? :) Naming is hard lol

harenson: haha

PR #175, opened 2021-04-12
KCRO-97: Add metadata on response when fetch, update, or delete db doc

harenson:
Depends on https://github.com/2600hz/kazoo-core/pull/251

All the APIs that return a database object use `crossbar_doc:handle_datamgr_success/2` for building the response within the given context, and because of that Kazoo can add the `object_metadata` key/value pair on the context on crossbar_doc:handle_datamgr_success/2 function, that way the metadata object can be added later on the API response on `api_util:create_default_resp_envelope/1` function before replying to the user.

The only API that was not using crossbar_doc:handle_datamgr_success/2 for building the response was phone_numbers API but that was also changed on this PR. Also, since this API doesn't use crossbar_doc for loading the information from the database it builds a custom response and the information required for building the metadata object is not on the same place (don't have the same field names) but this exception is being handled by `kz_doc:metadata/1` function.


src/api_util.erl
```diff
@@ -1750,7 +1754,7 @@ create_default_resp_envelope(Context) ->
                    ,{<<"request_id">>, cb_context:req_id(Context)}
                    ,{<<"revision">>, kz_term:to_api_binary(cb_context:resp_etag(Context))}
                    ,{<<"data">>, RespData}
-                   | BaseResp
+                   | maybe_add_metadata(Context, BaseResp)
```
icehess: If I understand this correctly, this always adds the metadata to the response if Context has `object_metadata` at the `metadata` key of resp JObj? like `{"data":{"metadata": {"modified": 123, "created": 123}}}`? 
Why not adding to `_read_only`?

icehess: nvm, thought it is adding to `data` object. but it is in resp root object.

harenson: The request is to add the **metadata** information outside of the **data** object.

harenson: To be explicit, is it not reading the `object_metadata` from the "resp JObj" but from the context storage and adds it to the root level of the API response; **metadata** object will be at the same level of **data** object.

jamesaimonetti (commented):
My opinion is KAZOO already uses `_read_only` in many places to include data about the doc that is not writable by the client. I would prefer to continue to use `_read_only` in the response doc over `metadata` as a top-level key in the response.

One forward-thinking reason too is we could start including `_read_only` in view results (when include_docs is used) as well.

@k-anderson looking for your thoughts here since you filed the ticket

harenson:
@jamesaimonetti KCRO-97 PRs are already using `_read_only` key to include the object's metadata. They are ready for another round of reviews in case you have time. This PR is failing because it depends on the changes on kazoo-core PR.

PR #190, opened 2021-06-21
PISTON-1289:create caller id if undefined

bradfordben:
Added new config attribute to disable kazoo from auto creating an users caller id name from the first and last name if no
caller id name was set. The default value is set to true so that it wont
change current behavior. This allows kazoo to use the UE caller id if
its set on the call leg.

Other required prs: https://github.com/2600hz/kazoo-core/pull/269

Kazoo 4.3: https://github.com/2600hz/kazoo/pull/6722


lazedo:
@bradfordben i really don't understand this use case, things are a bit changed in 5.x. can you provide 5.x logs of before and after with kazoo/fs logs ? thanks.

bradfordben:
@lazedo Yes i can do that. It will take me abit of time to get everything set up bit il get them for you this week.

The use case is for when a sip client wants to set the caller id on the device as opposed to it being set in kazoo. So at the moment if the endpoint has no caller id name in the caller_id object then the name is generated from the first and last name of the user. This is just a fix to allow kazoo to not overwrite the caller id and use the caller id from the A leg. Il verify its needed in kazoo5 but i would expect it is as the code is very similar around this.

priv/couchdb/schemas/users.json
```diff
@@ -53,6 +53,11 @@
             },
             "type": "object"
         },
+        "create_caller_id_name_if_undefined": {
+            "description": "Create the users caller id name from the first and last name fields if no caller id is defined",
+            "support_level": "supported",
```
jamesaimonetti: Just a note: this key is for 2600Hz-supported features / flags only.

Also, for boolean-returning flags, a "should_" prefix is usually preferred: `should_create_caller_id` would probably suffice.

jamesaimonetti (changes requested):

priv/couchdb/schemas/users.json
bradfordben: Ah ok i never knew what the supported was for 👍  Sounds good il remove it. 

`should_` sounds good. il make that change.

bradfordben: Changes made

lazedo:
@bradfordben i think this doesn't apply to 5.x

jamesaimonetti:
@bradfordben just going through older PRs to see what's still relevant. Can you re-open the PR with appropriate changes if you evaluate your usecase against 5.0 and find changes are still needed?

Thanks!

PR #191, opened 2021-06-21
KZOO-158: Add rename_and_refresh_app functionality

harenson:
After renaming an application's folder, this functionality can be used to rename the application on all the accounts using it.


src/crossbar_maintenance.erl
```diff
@@ -774,6 +775,14 @@ refresh_app(AppPath) ->
 refresh_app(AppPath, AppUrl) ->
     kazoo_documents_maintenance:refresh_app(AppPath, AppUrl).
 
+-spec rename_and_refresh_app(file:filename_all(), kz_term:ne_binary()) -> 'ok'.
+rename_and_refresh_app(AppPath, PrevName) ->
+    rename_and_refresh_app(AppPath, PrevName, 'undefined').
+
+-spec rename_and_refresh_app(file:filename_all(), kz_term:ne_binary(), kz_term:api_binary()) -> 'ok'.
+rename_and_refresh_app(AppPath, PrevName, AppUrl) ->
+    kazoo_documents_maintenance:rename_and_refresh_app(AppPath, PrevName, AppUrl).
```
icehess, line 794: I feel like it is better either implement the whole thing here instead of core or remove this redundant code.

harenson: I did it that way because that is the way all the other functions work within this module, they are just a link to functions with the same name within `kazoo_documents_maintenance` module. I think you were the one that moved all the functions' body out of crossbar_maintenance module to kazoo_documents_maintenance module and then left the functions on crossbar_maintenance module as a callback I guess and backwards compatibility.

I tried to move all the changes for this particular ticket to crossbar_maintenance, but the code relays on several private functions that were moved from crossbar_maintenance to kazoo_documents_maintenance module in a previous PR.

PR #193, opened 2021-06-29
PROD-199: create a new API to manage and verify external numbers

k-anderson:
Created a new API to add/remove external numbers as well as verify that the the user adding the number is also able to receive calls at that number.


k-anderson:
Related PRs:
https://github.com/2600hz/kazoo-core/pull/274
https://github.com/2600hz/kazoo-stepswitch/pull/18
https://github.com/2600hz/kazoo-ecallmgr/pull/94
https://github.com/2600hz/kazoo-crossbar/pull/193

src/modules/cb_external_numbers.erl
```diff
@@ -0,0 +1,499 @@
...
+                ,{<<"Account-Realm">>, Realm}
+                ,{<<"To-DID">>, Number}
+                ,{<<"Outbound-Caller-ID-Name">>, CIDName}
+                ,{<<"Outbound-Caller-ID-Number">>, CIDNumber}
+                ,{<<"Resource-Type">>, <<"originate">>}
+                ,{<<"Application-Name">>, <<"park">>}
+                ,{<<"Originate-Immediate">>, 'true'}
+                ,{<<"Timeout">>, 30}
+                | kz_api:default_headers(?APP_NAME, ?APP_VERSION)
+                ]),
```
lazedo, line 470: related to the stepswitch related PR
we should include here the CCVs for authorization `if they are really needed`, the stepswitch `CCVUpdates` in related pr should be removed.

lazedo (commented):
determine if/why authorization is needed

src/modules/cb_external_numbers.erl
k-anderson: Ya, I could do that... landed here because the stepswitch originate creates an A-leg programmatically so had a feeling that it would always be a "device-less/user-less" leg since its not built with any endpoint parameters and such to simplify just standardize as an account leg... flipped a bunch on this so no biggie.

src/modules/cb_external_numbers.erl
```diff
@@ -0,0 +1,499 @@
...
+    Routines = [fun validate_number_reconcilable/1
+               ,fun validate_number_unique/1
+               ],
+    lists:foldl(fun(F, C) -> F(C) end, initalize_fields(Context), Routines).
+
+-spec initalize_fields(cb_context:context()) -> cb_context:context().
+initalize_fields(Context) ->
+    Doc = cb_context:doc(Context),
+    Number = knm_converters:normalize(
+               kzd_external_numbers:number(Doc)
```
jamesaimonetti, line 227: https://github.com/2600hz/kazoo-core/pull/274/files#diff-b98c4b35703f7d70a0e0d0d86df9333e26920142e9072d1dd845d49ff7e71c9cR116 says `number/1` returns an `api_object()` and `knm_converters:normalize/1` takes binary or binaries.

So is the schema/accessor wrongly speced?

src/modules/cb_external_numbers.erl
```diff
@@ -0,0 +1,499 @@
...
+%% going to be responded to.
+%% @end
+%%------------------------------------------------------------------------------
+
+-spec allowed_methods() -> http_methods().
+allowed_methods() ->
+    [?HTTP_GET, ?HTTP_PUT].
+
+-spec allowed_methods(path_token()) -> http_methods().
+allowed_methods(_Thing) ->
```
jamesaimonetti: Would prefer `_NumberId` or `_SomethingMeaningful` to `_Thing`. These variable names are used in schema generation and stuff.

src/modules/cb_external_numbers.erl
```diff
@@ -0,0 +1,499 @@
...
+    case place_call(Context) of
+        {'ok', JObj} ->
+            play_code(Context, kz_json:get_value(<<"Call">>, JObj));
+        {'timeout', _} ->
+            cb_context:add_system_error(<<"call failure">>, Context);
+        {'error', E} ->
+            cb_context:add_system_error(E, Context)
+    end.
+
+-spec play_code(kapps_call:call(), kz_json:object()) -> 'ok'.
```
jamesaimonetti: 2nd arg should be `kz_term:api_object()`

src/modules/cb_external_numbers.erl
```diff
@@ -0,0 +1,499 @@
...
+    case place_call(Context) of
+        {'ok', JObj} ->
+            execute_ivr(Context, kz_json:get_value(<<"Call">>, JObj));
+        {'timeout', _} ->
+            cb_context:add_system_error(<<"call failure">>, Context);
+        {'error', E} ->
+            cb_context:add_system_error(E, Context)
+    end.
+
+-spec execute_ivr(kapps_call:call(), kz_json:object()) -> 'ok'.
```
jamesaimonetti: api_object here too

src/modules/cb_external_numbers.erl
k-anderson: I already did but haven't been able to find the correct rebuild command to update the swagger/docs so didn't commit yet... happen to know (reset priv/api and priv/oas3 to 5.0, ran make apis;make docs;make schemas;make code_checks but didn't rebuild anything so now its just the reset removing what was there)? :)

src/modules/cb_external_numbers.erl
k-anderson: returns api_ne_binary, updated core PR

src/modules/cb_external_numbers.erl
lazedo: i think its not needed as the stepswitch pr is ok

jamesaimonetti:
Added https://github.com/2600hz/kazoo-ast/pull/13 to address schema thing

src/modules/cb_external_numbers.erl
```diff
@@ -0,0 +1,507 @@
+%%%-----------------------------------------------------------------------------
+%%% @copyright (C) 2011-2021, 2600Hz
+%%% @doc Crossnar API for skel.
```
lazedo:
```suggestion
%%% @doc Crossbar API for external numbers.
```

PR #196, opened 2021-06-30
KCRO-93: Extend query_channels response validation function

harenson:
And also federate the request in order to get responses from ecallmgr on other zones in case it is necessary. i.e.: when the `GET /channels/UUID` request hits a zone that knows nothing about the given channel.


src/modules/cb_channels.erl
```diff
@@ -21,6 +21,9 @@
         ,put/2
         ]).
 
+%% Exported only because it is a callback for kz_amqp_worker to validate a response.
+-export([custom_query_channels_resp_v/1]).
+
```
lazedo: not needed

src/modules/cb_channels.erl
```diff
@@ -183,9 +186,16 @@ channels_query(CallId) ->
 
     kz_amqp_worker:call_collect(Req
                                ,fun kapi_call:publish_query_channels_req/1
-                               ,{'ecallmgr', fun kapi_call:query_channels_resp_v/1}
+                               ,{'ecallmgr', fun ?MODULE:custom_query_channels_resp_v/1, 'true'}
```
lazedo: not need for `?MODULE:`

lazedo (commented):
looks good

harenson:
Thanks for reviewing 👍🏾

PR #198, opened 2021-07-02
Remove users from directories schema

jamesaimonetti:
A user's doc is updated under the "directories" key to add them to a
directory listing, not on the directory document itself.


jamesaimonetti:
Needs and wants https://github.com/2600hz/kazoo-core/pull/282

PR #199, opened 2021-07-08
KCRO-104: Avoid conferences public links to expire

harenson:


lazedo:
@harenson you can do this with configuration. no need for this change. the spell check changes are welcome :)

icehess:
@lazedo we can;t do this with configuration, at least without any code change. if we remove expiration from auth module config crossbar-auth still set a default one. Otherwise I agree that this can be done using config.

lazedo:
@icehess i believe you can, just configure the module to not set an expiration

harenson:
@lazedo Do you mean setting `token_auth_expiry_s = undefined`? If so, it will be caught by this match_all rule and it will use the default value as @icehess said https://github.com/2600hz/kazoo-crossbar/blob/master/src/crossbar_auth.erl#L241-L247

lazedo:
did you try to set  `token_auth_expiry_s` to 0  for the method (module name) ?

harenson:
It will end up on the same catch_all branch of the case because the
expiration time is not greater than (>) 0.

lazedo:
so it will be `undefined` which is what you're after. try it

harenson:
The only possible way to get an `undefined` value with the current code for generating the expiration time is if `token_auth_expire_s` and `?DEFAULT_AUTH_EXPIRY` are either `0` or `undefined`. If I set `token_auth_expire_s = 0` and `?DEFAULT_AUTH_EXPIRY` is different than zero or undefined, that value will be used, that's the default value.

I tried as you suggested setting the module's settings with the `token_auth_expire_s` value set to zero and as expected it is using the default value:
```erlang
{<<"cb_conference_auth">>,
 {[{<<"enabled">>,true},
   {<<"token_auth_expiry_s">>,0},
   {<<"log_failed_attempts">>,true},
   {<<"log_successful_attempts">>,false},
   {<<"scopes">>,[<<"crossbar:conference_join">>]}]}
                 
Now: 1626302084, Expiration: 1626305684 (3600s)
```

lazedo:
@harenson yep, i missed the `_ ->` which doesn't bind `TokenExp`, i went to look and its something i have locally.
```
    case kz_json:get_integer_value(Path, AuthConfig, undefined) of
        undefined ->
            case ?DEFAULT_AUTH_EXPIRY of
                TokenExp when TokenExp > 0 ->
                    kz_time:current_unix_tstamp() + TokenExp;
                _ -> 'undefined'
            end;
        TokenExp when TokenExp > 0 ->
            kz_time:current_unix_tstamp() + TokenExp;
        _Else -> undefined
    end.
```

harenson:
@lazedo Oh, I see. So @lazedo @icehess do you think this PR needs more work or is it ok as it is right now?

icehess:
@harenson Go ahead and do the change Luis showed in the previous comment (the case code above). Module's config can set the expiration to a negative number or `infinity` to disable the expiration. Also change the default value for `cb_conference_auth` to `infinity`.

Remove that `Option` change.

harenson:
@icehess done.

PR #203, opened 2021-07-26
PISTON-1267 add a handler for action 'move' to channel's crossbar API

alexey-khabulyak:
Add ability to move a call(call-flipping) via crossbar API
4.3 PR: https://github.com/2600hz/kazoo/pull/6728


src/modules/cb_channels.erl
```diff
@@ -545,6 +547,17 @@ maybe_intercept(Context, CallId, TargetType, TargetId) ->
     _ = kz_amqp_worker:cast(API, fun kapi_metaflow:publish_action/1),
     crossbar_util:response_202(<<"intercept initiated">>, Context).
 
+-spec maybe_move(cb_context:context(), kz_term:ne_binary()) -> cb_context:context().
+maybe_move(Context, CallId) ->
+    API = [{<<"Call-ID">>, CallId}
+          ,{<<"Action">>, <<"move">>}
+          ,{<<"Data">>, kz_json:new()}
```
jamesaimonetti: Data is not required, seems unnecessary to include an empty JSON object?

jamesaimonetti: Looking more, it seems like `Data` would want some data: https://docs.2600hz.com/dev/applications/konami/doc/move/

I assume you'll want to populate from the API request data

alexey-khabulyak: @jamesaimonetti hi.
I've added Data because actions "Hangup" and "Break" also contain empty data. So I assumed that move also needs it.

Also, i've added some Data like this:
```
    API = [{<<"Call-ID">>, CallId}
          ,{<<"Action">>, <<"move">>}
          ,{<<"Data">>, kz_json:from_list([{<<"auto_answer">>, cb_context:req_value(Context, <<"auto_answer">>, 'false')}
                                          ,{<<"can_call_self">>, cb_context:req_value(Context, <<"can_call_self">>, 'true')}
                                          ,{<<"device_id">>, cb_context:req_value(Context, <<"device_id">>)}
                                          ,{<<"dial_strategy">>, cb_context:req_value(Context, <<"dial_strategy">>, <<"simultaneous">>)}
                                          ,{<<"owner_id">>, cb_context:req_value(Context, <<"owner_id">>)}
                                          ])}
           | kz_api:default_headers(?APP_NAME, ?APP_VERSION)
          ],
```
 
But seems like `can_call_self` doesn't work. Can u look into the source code? Or I can add this param anyway?

PR #206, opened 2021-08-19
First Version of screenpops API

kdamas93:
-First Version of screenpops API


priv/couchdb/views/screenpops.json
```diff
@@ -0,0 +1,26 @@
...
+        ]
+    },
+    "language": "javascript",
+    "views": {
+        "config_listing": {
+            "map": [
+                "function(doc) {",
+                "  if (doc.pvt_type != 'screenpops_config' || doc.pvt_deleted) return;",
+                "  emit(doc._id, {",
+                "    'id': doc._id,",
```
jamesaimonetti: id is included on all view results, no need to include it in the view as well. You'll get "key", "value" (the emitted object here), and "id" I believe.

src/modules/cb_screenpops.erl
```diff
@@ -0,0 +1,187 @@
...
+    validate_request(Context, cb_context:req_verb(Context)).
+validate_request(Context, ?HTTP_GET) ->
+    load_summary(Context);
+validate_request(Context, ?HTTP_PUT) ->
+    create_entry(Context).
+
+
+
+-spec validate(cb_context:context(), path_token()) -> cb_context:context().
+validate(Context, _DocId) ->
```
jamesaimonetti: Change to `_ScreenpopId` (and try not to use `_` in function args; we still like to see what the value represents). But in this case, tooling uses the variable name for doc generation.

kdamas93: Changed ! Ty.

priv/couchdb/views/screenpops.json
kdamas93: Id is needed when we do a get all

doc/screenpops.md
```diff
@@ -0,0 +1,56 @@
...
+
+#### About Screenpops
+
+#### Schema
+
+
+
+#### Fetch
+
+> GET /v2/accounts/{ACCOUNT_ID}/screenpops
```
jamesaimonetti, line 35: Can you include expected JSON responses for all requests and sample JSON request data for PUT/POST/etc?

priv/couchdb/schemas/screenpops_config.json
```diff
@@ -0,0 +1,45 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "_id": "screenpops_config",
```
jamesaimonetti: I question why `screenpops_config` vs `screenpops`?

For most of the API schemas, if the endpoint is `users` then the schema matches.

jamesaimonetti: Would also help to get a `kzd_screenpops.erl` accessor module in `core/kazoo_documents`

priv/oas3/oas3-parameters.yml
```diff
@@ -181,6 +181,16 @@
     'minLength': 32
     'pattern': '^[0-9a-f]+$'
     'type': string
+'DOC_ID':
```
jamesaimonetti: remove this so it can get re-generated with `SCREENPOP_ID` or whatever the tooling decides

priv/oas3/openapi.yml
```diff
@@ -494,6 +494,10 @@
   '/accounts/{ACCOUNT_ID}/scope_restrictions/{SCOPE_RESTRICTION}':
     '$ref': >-
       paths/scope_restrictions.yml#/paths/~1accounts~1{ACCOUNT_ID}~1scope_restrictions~1{SCOPE_RESTRICTION}
+  '/accounts/{ACCOUNT_ID}/screenpops':
+    '$ref': 'paths/screenpops.yml#/paths/~1accounts~1{ACCOUNT_ID}~1screenpops'
+  '/accounts/{ACCOUNT_ID}/screenpops/{DOC_ID}':
```
jamesaimonetti: Same here

src/modules/cb_screenpops.erl
```diff
@@ -0,0 +1,187 @@
...
+    cb_context:validate_request_data(<<"screenpops_config">>, Context, OnSuccess).
+
+%%------------------------------------------------------------------------------
+%% @doc
+%% @end
+%%------------------------------------------------------------------------------
+-spec on_successful_validation(kz_term:api_binary(), cb_context:context()) ->
+          cb_context:context().
+on_successful_validation('undefined', Context) ->
+    cb_context:set_doc(Context, kz_json:set_values([{<<"pvt_type">>, <<"screenpops_config">>}
```
jamesaimonetti: `pvt_type` should probably be `screenpop` to match project standards.

kdamas93: changed, tyvm for the explanation

src/modules/cb_screenpops.erl
```diff
@@ -0,0 +1,187 @@
...
+%% going to be responded to.
+%% @end
+%%------------------------------------------------------------------------------
+
+-spec allowed_methods() -> http_methods().
+allowed_methods() ->
+    [?HTTP_GET, ?HTTP_PUT].
+
+-spec allowed_methods(path_token()) -> http_methods().
+allowed_methods(_DocId) ->
```
jamesaimonetti: I think this is causing `DocId` in the docs to be generated. Make sure all instances in the code are changed to `_ScreenpopId` or `ScreenpopId` as needed.

kdamas93: Changed! Tyvm

doc/screenpops.md
```diff
@@ -0,0 +1,237 @@
+### Screenpops
+
+#### About Screenpops
```
jamesaimonetti, line 3: Can you add some text about what this API would be used for? Also, I think you need to rename the schema from `screenpops_config.json` to `screenpops.json`. You should then get a markdown table with the schema inserted into the doc (and ref doc).

priv/couchdb/schemas/screenpops_config.json
```diff
@@ -0,0 +1,50 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "_id": "screenpops",
+    "properties": {
+        "custom_parameters": {
+            "properties": {
+                "custom_url": {
+                    "type": "string"
```
jamesaimonetti: Add basic `"description":"what is this field even?"` properties

jamesaimonetti (changes requested):

jamesaimonetti:
One more thing - the schema will generate a file in `/core/kazoo_documents/` that will need its own PR. Copy the `screenpop.erl.src` file to `screenpop.erl` and flesh out any functions needed.

doc/screenpops.md
kdamas93: Explanation added and file renamed

priv/couchdb/schemas/screenpops_config.json
kdamas93: Description added

jamesaimonetti:
See:
- https://github.com/2600hz/kazoo-core/pull/296
- https://github.com/2600hz/kazoo-properly/pull/63

PR #214, opened 2021-09-17
KCRO-115: user info auth resp

icehess:
- add basic user info to auth response
- fix auth/authrozied doesn't have schema


src/crossbar_doc.erl
```diff
@@ -114,7 +116,7 @@ load(DocId, Context, Options, _RespStatus) when is_binary(DocId) ->
             case check_document_type(Context, JObj, Options) of
                 'true' ->
                     JObj1 = maybe_normalize_doc(JObj, props:get_value('doc_normalizer', Options)),
-                    handle_successful_load(Context, JObj1, Options);
+                    handle_successful_load(Context, has_avatar(JObj1), Options);
```
icehess: Why adding this in `crossbar_doc`? This change should be in `cb_users:load_user/2`.

icehess: change `cb_users:load_user/2` to:
```erlang
load_user(UserId, Context) ->
    Context1 = crossbar_doc:load(UserId, Context, ?TYPE_CHECK_OPTION(kzd_users:type())),
    case cb_context:resp_status(Context1) of
        'success' -> add_has_avatar(Context1);
        _ -> Context1
    end.
```

For checking the doc has the attachment you can also use `kz_doc:attachment(Doc, ?PHOTO)`.

oguerrero913: you are right, moved in new commit

PR #216, opened 2021-09-21
Fixed chunked encoding of potential json terms.

lordnull:
This was discovered when the `page_size` property for cdr's was sometimes
a string, and some times a number.

When finishing a chunked return using json, the trailer properties were
encoded by first turning the property values into binaries, and then
wrapping those binaries in quotes. This meant the erlang term `43` always
become `"43"`.

By checking if the erlang term is already a json term, and only turning to
a binary if it is not, then we can avoid encoding things like json objects,
`null`, booleans, and numbers into strings, and things like tuples, pids
(and more) are still handled as before.

The spec doesn't fully reflect the above, however, as we don't yet have
anything other than strings and numbers.

4.3 pr: https://github.com/2600hz/kazoo/pull/6739


src/api_util.erl
```diff
@@ -1774,10 +1774,7 @@ close_chunk_json_envelope(Req, Context) ->
         ,{<<"revision">>, kz_term:to_api_binary(cb_context:resp_etag(Context))}
         ,{<<"auth_token">>, cb_context:auth_token(Context)}
         ],
-    Encoded = [<<"\"", K/binary, "\":\"", (kz_term:to_binary(V))/binary, "\"">>
-                   || {K, V} <- Trailer,
-                      kz_term:is_not_empty(V)
-              ],
+    Encoded = encode_traier(Trailer),
```
jamesaimonetti: spellcheck?

src/api_util.erl
```diff
@@ -1787,6 +1784,35 @@ maybe_delete_range_keys(Resp, 'false') ->
...
+%%------------------------------------------------------------------------------
+%% @doc Encode a term to the closest json representation it has. The tightness
+%% of the spec is because this is used solely in `encode_trailer/1', so we're
+%% only getting binaries and numbers.
+%% @end
+%%------------------------------------------------------------------------------
+-spec encode_json_naively(kz_term:ne_binary() | number()) -> kz_json:json_term().
+encode_json_naively(Term) ->
+    case kz_json:is_json_term(Term) of
+        false ->
```
jamesaimonetti: tickie atoms please

jamesaimonetti:
@lordnull a couple things:

1. This is not a "bug", per se. I don't think the 4.3 PR will be merged as it is closed for non-bug contributions.
2. Please open a PR against master (all work should start with a master PR).
3. Is there a way to elicit the stringified `page_size` response? Would make a good contribution to kazoo-properly to check that the response always has a number where expected

lordnull:
1. Okay. I'm new, so I'll want to run this by my onboarder before I just close the pr, I'll just take no action on it.
2. I've rebased the code-change and repointed the pr as requested.
3. Adding `is_chunked=true` in the query parameters adds the page size for basically any endpoint that has multiple rows of data. Without the fix, the page_size is always a string. With it, the page_size is an integer. I'll look into kazoo-properly to get a feel on how to add the check(s).

jamesaimonetti:
@lordnull first, welcome! Glad to have new eyes on the code. Please speak up about things that don't seem right - its either something we've grown accustom to and should change, or it has a non-obvious reason that probably requires improved docs.

For point 2, the typical flow is get a master PR reviewed / merged, then backport to release branch(es) if required. As we're pushing for stable 5.0, and non-bug fixes aren't backported, a 5.0 PR likely isn't necessary. If you need it on a 5.0 release, then we take a closer look and coordinate with QA about making the backport. As a hypothetical, if this PR is required by you to be backported, we would likely do so as the PR doesn't materially impact things (at least as far as our CI testing goes - QA's suite may trigger an issue but then we'd have that conversation).

For point 3, I would look at [seq_cdrs](https://github.com/2600hz/kazoo-properly/blob/master/src/seq/seq_cdrs.erl) and add a test that generates the CDRs, lists them via the summary API with `is_chunked`, and use `kz_json:get_value/2` to get `page_size` - it should be an integer (vs using the type-casting `kz_json:get_integer_value/2`. Be sure to test before the patch is applied to make sure the test fails (decodes to a binary) and that the test passes after the patch.

Thanks! Let me know if you have further questions.

PR #218, opened 2021-09-29
KZOO-205: Fix ACL ip extraction when building CIDRs

harenson:
5.0: https://github.com/2600hz/kazoo-crossbar/pull/233

Before this change, if an ACL had several IPs on its CIDR value, `cb_devices_utils:extract_ip/3` function would treat that list of IPs as a single IP and would also try to build a single CIDR with all those IPs as the value.


src/cb_devices_utils.erl
```diff
@@ -79,10 +79,10 @@ extract_all_ips(JObj) ->
 extract_ip(_Key, Value, Acc) ->
     case kz_json:get_value(<<"cidr">>, Value) of
         'undefined' -> Acc;
-        CIDR ->
-            [kz_json:from_list([{<<"ip">>, CIDR}
+        CIDRs ->
```
jamesaimonetti, line 82: Should use get_list_value to show the typed nature of the value

PR #219, opened 2021-09-29
Changes related to bumped deps

jamesaimonetti:


jamesaimonetti:
Requires https://github.com/2600hz/kazoo5/pull/64

PR #221, opened 2021-10-12
Adding endpoint to generate token to login into getstreamio API

kdamas93:


src/modules/cb_getstream.erl
```diff
@@ -0,0 +1,144 @@
...
+    case kzd_accounts:fetch(ResellerId) of
+        {'ok', JObj} -> kz_json:get_ne_binaries([<<"getstream">>, <<"secret">>], JObj);
+        {'error', _R} ->
+            lager:debug("unable to get reseller document ~s: ~p", [ResellerId, _R]),
+            'undefined'
+    end.
+
+-spec maybe_get_default_secret(cb_context:context()) -> kz_term:api_binary().
+maybe_get_default_secret(_Context) ->
+    kapps_config:get_ne_binary(<<(?CONFIG_CAT)/binary, ".getstream">>, <<"secret">>, <<"">>).
```
icehess: Remove `<<"">>`, otherwise it never returns `undefined` and we always create JWT with an empty secret.

src/modules/cb_getstream.erl
```diff
@@ -0,0 +1,144 @@
...
+put(Context, ?TOKEN_PATH) ->
+    UserData = cb_context:doc(Context),
+    UserId = kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>],<<"username">>],UserData),
+    Secret = cb_context:fetch(Context, 'getstream_secret'),
+    TokenId = kz_datamgr:get_uuid(),
+    {_, Token} = kz_auth_jwt:encode(
+                   #{claims => [{<<"user_id">>, UserId}],
+                     issuer => ?ISSUER,
+                     alg => ?ALG,
+                     key => Secret}
```
icehess, line 101: looking at `kz)auth_jwt` it seems key is a Private Key. Is the secret is in form of Pem? you also need to decode it. When signing the JWT it uses `public_key:sign/3` which is expecting a decoded private key.

icehess (changes requested):
Left some comments regarding secret/private_key.

Another question, what is this `TokenId` use for?

src/modules/cb_getstream.erl
kdamas93: That's the intention to create a placeholder, this secret key will be added manually or by cluster manager

src/modules/cb_getstream.erl
kdamas93: The key that will be saved in the doc is already decode, is the way that getsream sent to us, I did tests passing that key to sign/3, the connection was successful.
There is already a PR related to this implementation : https://github.com/2600hz/kazoo-core/pull/311
We will use crypto:mac because getstream works with hs256 alg.

kdamas93:
> Left some comments regarding secret/private_key.
> 
> Another question, what is this `TokenId` use for?

Is been using just to be located in the response

jamesaimonetti:
@kdamas93 you should run `make ci-codechecks` from the kazoo root.

Missing files include:
- `doc/ref/getstream.md` which is the auto-generated API doc page
- `priv/couchdb/schemas/system_config.crossbar.getstream.json` - the schema file for the system config doc
- `priv/oas3/paths/getstream.yml`

I suspect CI checks aren't errors on untracked files generated. Will check why this PR is considered passing in that case.

kdamas93:
Files added ! Thanks for your help @jamesaimonetti

doc/ref/getstream.md
```diff
@@ -0,0 +1,18 @@
...
+
+## About Getstream
+
+## Schema
+
+
+
+## Create
+
+> PUT /v2/accounts/{ACCOUNT_ID}/getstream/token
```
jamesaimonetti, line 11: Can you sync up the docs to match. Headings should start with a single `#` then nest further, for instance.

doc/getstream.md
```diff
@@ -0,0 +1,31 @@
+### Getstream
+
+#### About Getstream Token
```
jamesaimonetti: Please match headings to the ref doc (this should be `##` for instance)

PR #222, opened 2021-10-19
Added API

oguerrero913:
do not merge

Added API
- add/get/get all clusters
- add/get/get all apps

 working on
- add/get/get all ownership
- patch/delete/post cluster/apps/ownership


icehess:
@oguerrero913 Please create PR in https://github.com/2600hz/kazoo-appex-server repo instead and close this.

PR #224, opened 2021-10-27
Set default claims onto request's claims object

jamesaimonetti:
Before, if the request's claims had an identical pair (say account_id)
it would result in a duplicated key, throwing an error in kz_json.

This ensures crossbar_auth's keys are set over the request's claims if
there's duplicated keys.


icehess (approved):

src/crossbar_auth.erl
```diff
@@ -78,15 +78,16 @@ create_auth_token(Context, Method, JObj) ->
...
-               | kz_json:to_proplist(kz_json:get_value(<<"Claims">>, JObj, kz_json:new()))
-               ]),
+    ReqClaims = kz_json:to_proplist(kz_json:get_value(<<"Claims">>, JObj, kz_json:new())),
+    Claims = props:set_values([{<<"account_id">>, AccountId}
+                              ,{<<"owner_id">>, OwnerId}
+                              ,{<<"as">>, kz_json:get_value(<<"as">>, Data)}
+                              ,{<<"method">>, Method}
+                              ,{<<"exp">>, Expiration}
+                              ,{<<"mfa_resp">>, kz_json:get_ne_value(<<"multi_factor_response">>, Data)}
+                              ]
```
lazedo: should filter `undefined` first then set values, orelse  these will be set as undefined in claims ?

src/crossbar_auth.erl
```diff
@@ -78,15 +78,15 @@ create_auth_token(Context, Method, JObj) ->
...
-               ]),
+    ReqClaims = kz_json:to_proplist(kz_json:get_json_value(<<"Claims">>, JObj, kz_json:new())),
+    AdditionalClaims = props:filter_undefined([{<<"account_id">>, AccountId}
+                                              ,{<<"owner_id">>, OwnerId}
+                                              ,{<<"as">>, kz_json:get_value(<<"as">>, Data)}
+                                              ,{<<"method">>, Method}
+                                              ,{<<"exp">>, Expiration}
+                                              ,{<<"mfa_resp">>, kz_json:get_ne_value(<<"multi_factor_response">>, Data)}
+                                              ]),
+    Claims = props:set_values(AdditionalClaims, ReqClaims),
```
lazedo, line 87: i think this should be inverted, ReqClaims on top of Additional(default) ?

lazedo: at least for `owner_id`

PR #225, opened 2021-11-05
[5.0] Adding system dialplan

filipevarjao:
Kazoo [docs](https://docs.2600hz.com/supported/doc/internationalization/numbers/#adding-system-dialplan-example) says we can add system dialplan by
`curl -X PUT -H "Content-Type: application/json" -H "X-Auth-Token: {AUTH_TOKEN}" http://server.com:8000/v2/system_configs -d '{"data":{"id":"dialplans"}}'
`
but we are responding it with 405 invalid_method


src/modules/cb_system_configs.erl
```diff
@@ -76,7 +76,7 @@ authorize(Context, _Id, _Node) -> authorize(Context).
 
 -spec allowed_methods() -> http_methods().
 allowed_methods() ->
-    [?HTTP_GET].
+    [?HTTP_GET, ?HTTP_PUT].
```
jamesaimonetti, line 79: Isn't this `PUT /v2/system_configs` vs `PUT /v2/system_configs/dialplan` ?

jamesaimonetti: I think the fix is to correct the docs to `PUT /v2/system_configs/dialplan -d '{"data":{"whatever":"data"}}'` no?

filipevarjao: now I have the changes to make it work as the docs description, I'll add test coverage then we'll see

filipevarjao:
going to open a new PR against master

PR #227, opened 2021-11-11
KCRO-129: Allow presence state changes only providing extension number

harenson:
Requires: https://github.com/2600hz/kazoo-core/pull/341

Right now, `/presence` API only allows to change presence state for a particular presentity if either users/{USER_ID} or devices/{DEVICE_ID} is provided. With these changes users can change presence state by only providing the extension number which was only allowed when user wanted to reset presence state.


src/modules/cb_presence.erl
```diff
@@ -307,7 +326,7 @@ invalid_action(Context) ->
     cb_context:add_validation_error(<<"action">>
                                    ,<<"invalid">>
                                    ,kz_json:from_list(
-                                      [{<<"message">>, <<"Field must be set to a valid action">>}
+                                      [{<<"message">>, <<"Field must be set to a valid action: set, reset">>}
```
jamesaimonetti: Can you take a look at how `enum` errors are returned; this seems like an `enum` now

src/modules/cb_presence.erl
```diff
@@ -319,7 +338,7 @@ invalid_state(Context) ->
     cb_context:add_validation_error(<<"state">>
                                    ,<<"invalid">>
                                    ,kz_json:from_list(
-                                      [{<<"message">>, <<"Field must be set to a valid state">>}
+                                      [{<<"message">>, <<"Field must be set to a valid state: ", (kz_binary:join(?PRESENCE_STATES, <<", ">>))/binary>>}
```
jamesaimonetti: Same, see `enum` errors

src/modules/cb_presence.erl
```diff
@@ -372,7 +397,9 @@ publish_presence_reset(Context, PresenceId) ->
     kz_amqp_worker:cast(API, fun kapi_presence:publish_reset/1).
 
 -spec publish_presence_update(cb_context:context(), kz_term:api_binary(), kz_term:ne_binary()) -> 'ok'.
-publish_presence_update(_Context, 'undefined', _PresenceState) -> 'ok';
+publish_presence_update(_Context, 'undefined', _PresenceState) ->
+    lager:debug("'undefined' presence_id provided, skipping presence state update"),
```
jamesaimonetti: no need to return `ok` explicitly

src/modules/cb_presence.erl
harenson: Do you have a link to an example?

harenson: Do you mean something like this https://github.com/2600hz/kazoo-crossbar/blob/master/src/modules/cb_devices.erl#L322-L333 ?

jamesaimonetti: maybe like https://github.com/2600hz/kazoo-crossbar/blob/b747e555a7c5ca32168b2183a647bc00b656a4fb/src/modules/cb_search.erl#L210

harenson:
Sorry about the forced push, I rebased master, thus I had to force push.

lazedo:
@harenson why are you removing the schemas ? shouldn't the `improve` nature of the `pr` actually use them to validate the requests ?

harenson:
@lazedo I tried to do validation using schemas but it was not working as expected and giving me some unexpected errors, also it was taking me too long and since that was not the changed @jamesaimonetti suggested on his comment I decided to remove those and do what actually James suggested/requested. But today I was able to make schemas work, so, there they are again and working this time :)

PR #229, opened 2021-11-11
[master] adding sustem_config dialplans endpoint

filipevarjao:
Kazoo [docs](https://docs.2600hz.com/supported/doc/internationalization/numbers/#adding-system-dialplan-example) says we can create system dialplans by
`curl -X POST -H "Content-Type: application/json" -H "X-Auth-Token: {AUTH_TOKEN}" http://server.com:8000/v2/system_configs/dialplans -d '{"data":{"^(2\\d{6})$":{"prefix":"+7383","name":"Novosibirsk"}}}'
`
but we are responding it with 405 invalid_method


src/modules/cb_system_configs.erl
```diff
@@ -141,6 +143,13 @@ validate_document(Context, Id, ?HTTP_GET) ->
 validate_document(Context, Id, ?HTTP_PUT) ->
     validate_document(Context, Id, ?HTTP_POST);
 
+validate_document(Context, ?DIALPLANS, ?HTTP_POST) ->
+    Dialplans = cb_context:req_data(Context),
+    [Key] = kz_json:get_keys(Dialplans),
```
jamesaimonetti, line 148: Doesn't this limit you to only one regex to match on? What if I wanted to add https://docs.2600hz.com/dev/doc/internationalization/examples/dialplan/NL_Amsterdam/ ?

lazedo:
@filipevarjao this `pr` is way off

PR #232, opened 2021-11-30
Adding functions to create/validate users in getstream

kdamas93:


doc/ref/getstream.md
```diff
@@ -8,11 +8,11 @@
...
 ## Create
 
-> PUT /v2/accounts/{ACCOUNT_ID}/getstream/token
+> PUT /v2/accounts/{ACCOUNT_ID}/getstream/connect
 
 ```shell
 curl -v -X PUT \
     -H "X-Auth-Token: {AUTH_TOKEN}" \
-    http://{SERVER}:8000/v2/accounts/{ACCOUNT_ID}/getstream/token
+    http://{SERVER}:8000/v2/accounts/{ACCOUNT_ID}/getstream/connect
```
jamesaimonetti, line 16: Changes to the ref/ doc should be reflected in the real doc

src/modules/cb_getstream.erl
```diff
@@ -139,6 +177,108 @@ maybe_get_reseller_secret(Context) ->
...
 -spec maybe_get_default_secret(cb_context:context()) -> kz_term:api_binary().
 maybe_get_default_secret(_Context) ->
     kapps_config:get_ne_binary(<<(?CONFIG_CAT)/binary, ".getstream">>, <<"secret">>).
+
+-spec maybe_get_default_api_key(cb_context:context()) -> kz_term:api_binary().
+maybe_get_default_api_key(_Context) ->
+    kapps_config:get_ne_binary(<<(?CONFIG_CAT)/binary, ".getstream">>, <<"api_key">>).
+
+-spec validate_if_user_exist_on_getstream( kz_term:api_binary() | 'undefined', cb_context:context(), kz_term:api_binary(), kz_term:api_binary(), kz_json:api_json_term()) -> boolean().
+validate_if_user_exist_on_getstream(UserId, Context, Secret, ApiKey, UserData ) ->
```
jamesaimonetti: If UserId is undefined, should a request to getstream be made? Or should it cause an error to be returned?

src/modules/cb_getstream.erl
```diff
@@ -139,6 +177,108 @@ maybe_get_reseller_secret(Context) ->
...
+
+    Eq = kz_json:from_list([{<<"$eq">>,UserId}]),
+    Id = kz_json:from_list([{<<"id">>, Eq}]),
+    Params = kz_json:from_list([{<<"filter_conditions">>, Id}]),
+    JsParams = kz_json:encode(Params),
+
+    case kz_http:get(BaseUrl ++ binary_to_list(kz_http_util:urlencode(JsParams)), Headers) of
+        {'ok', _Code, _RespHeaders, RespBody} ->
+            RespDecoded = kz_json:decode(RespBody),
+            check_getstream_body(kz_json:get_value(<<"users">>, RespDecoded), UserId, Context, ApiKey, Secret, UserData);
```
jamesaimonetti: What shape is the "users" value? Use the typed kz_json accessor (maybe `kz_json:get_ne_binaries/2`?) to access them.

src/modules/cb_getstream.erl
```diff
@@ -139,6 +177,108 @@ maybe_get_reseller_secret(Context) ->
...
+        _E ->
+            lager:debug("failed: ~p", [_E]),
+            'false'
+    end.
+
+-spec check_getstream_body(list(), kz_term:api_binary(), cb_context:context(), kz_term:api_binary(), kz_term:api_binary(), kz_json:api_json_term()) -> boolean().
```
jamesaimonetti: `list()` is too generic for a function specifically processing the items as JSON objects. `kz_json:objects()` would be a more helpful type.

src/modules/cb_getstream.erl
jamesaimonetti: Looking further, `kz_json:get_list_value/2` would fetch the value as a list of values.

src/modules/cb_getstream.erl
```diff
@@ -139,6 +177,108 @@ maybe_get_reseller_secret(Context) ->
...
+check_getstream_body([], UserId, Context, ApiKey, Secret, UserData) ->
+    create_getstream_user(UserId, Context, ApiKey, Secret, UserData);
+check_getstream_body([H|_T], _UserId, _Context, _ApiKey, _Secret, _UserData) ->
+    kz_json:get_value(<<"id">>, H),
```
jamesaimonetti: This is basically a no-op. Was there intent to use the ID? Also, `[H|_T]` aren't helpful variable names. `[User | _Users]` perhaps?

src/modules/cb_getstream.erl
```diff
@@ -139,6 +177,108 @@ maybe_get_reseller_secret(Context) ->
...
+
 -spec maybe_get_default_secret(cb_context:context()) -> kz_term:api_binary().
 maybe_get_default_secret(_Context) ->
     kapps_config:get_ne_binary(<<(?CONFIG_CAT)/binary, ".getstream">>, <<"secret">>).
+
+-spec maybe_get_default_api_key(cb_context:context()) -> kz_term:api_binary().
+maybe_get_default_api_key(_Context) ->
+    kapps_config:get_ne_binary(<<(?CONFIG_CAT)/binary, ".getstream">>, <<"api_key">>).
+
+-spec validate_if_user_exist_on_getstream( kz_term:api_binary() | 'undefined', cb_context:context(), kz_term:api_binary(), kz_term:api_binary(), kz_json:api_json_term()) -> boolean().
```
jamesaimonetti: Spec is wrong.
`kz_term:api_binary()` already includes `'undefined'` so no need to include it.
APIKey is a list (previously binary_to_list call) so not `api_binary()`.

src/modules/cb_getstream.erl
```diff
@@ -139,6 +177,108 @@ maybe_get_reseller_secret(Context) ->
...
+    end.
+
+-spec check_getstream_body(list(), kz_term:api_binary(), cb_context:context(), kz_term:api_binary(), kz_term:api_binary(), kz_json:api_json_term()) -> boolean().
+check_getstream_body([], UserId, Context, ApiKey, Secret, UserData) ->
+    create_getstream_user(UserId, Context, ApiKey, Secret, UserData);
+check_getstream_body([H|_T], _UserId, _Context, _ApiKey, _Secret, _UserData) ->
+    kz_json:get_value(<<"id">>, H),
+    'true'.
+
+-spec create_getstream_user(kz_term:api_binary(), cb_context:context(), kz_term:api_binary(), kz_term:api_binary(), kz_json:api_json_term()) -> boolean().
```
jamesaimonetti: Go through with dialyzer to fix these specs. From the root kazoo directory, `make dialyze-hard` will spout out all the discrepancies found.

src/modules/cb_getstream.erl
```diff
@@ -139,6 +177,108 @@ maybe_get_reseller_secret(Context) ->
...
+
+    FirstName = kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>],<<"first_name">>],UserData),
+    LastName = kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>],<<"last_name">>],UserData),
+
+    Body = kz_json:from_list(
+             [{<<"id">>, UserId},
+              {<<"teams">>, [cb_context:account_id(Context)]},
+              {<<"username">>, kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>],<<"username">>],UserData)},
+              {<<"name">>, binary_to_list(FirstName) ++ binary_to_list(LastName)},
+              {<<"role">>,kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>],<<"priv_level">>],UserData)}]),
```
jamesaimonetti: Drop trailing `])` to newline, so the `[]` for the proplist match

src/modules/cb_getstream.erl
```diff
@@ -139,6 +177,108 @@ maybe_get_reseller_secret(Context) ->
...
+
+-spec create_token_for_getstream(atom(), kz_term:api_binary(), kz_term:api_binary()) -> {'ok', kz_term:ne_binary()} | {'error', any()}.
+create_token_for_getstream('header', Value, Secret) ->
+    kz_auth_jwt:encode(
+      #{claims => [{<<"resource">>, <<"users">>}
+                  ,{<<"action">>, Value}
+                  ,{<<"feed_id">>, <<"*">>}],
+        issuer => ?ISSUER,
+        alg => ?ALG,
+        key => Secret}
```
jamesaimonetti: Drop `}` so `{}` for the map line up

src/modules/cb_getstream.erl
```diff
@@ -139,6 +177,108 @@ maybe_get_reseller_secret(Context) ->
...
+     );
+create_token_for_getstream(_Type, Value, Secret) ->
+    kz_auth_jwt:encode(
+      #{claims => [{<<"user_id">>, Value}],
+        issuer => ?ISSUER,
+        alg => ?ALG,
+        key => Secret}
+     ).
+
+-spec create_headers(kz_term:ne_binary()) -> kz_term:proplist().
```
jamesaimonetti: I think there's a `kz_http:headers()` type exported you can use here instead of proplist.

src/modules/cb_getstream.erl
kdamas93: The UserId is taken from the user document (<<"_id">>), Is possible get an 'undefined' ?

src/modules/cb_getstream.erl
kdamas93: "users" is a json object , I guess get_value is appropiate.

src/modules/cb_getstream.erl
jamesaimonetti: Should not be undefined if the user doc is fetched from the db.

src/modules/cb_getstream.erl
jamesaimonetti: `check_getstream_body` spec says its a list (of JSON objects), and usage in the function is one of a list of objects.

jamesaimonetti: And to be clear, if "users" was a JSON object, `kz_json:get_json_object/2` would be appropriate. `get_value/2` should only really be used if the field has multiple types for a value (like it could be an integer or a string).

doc/ref/getstream.md
jamesaimonetti: @kdamas93 can you make sure `doc/getstream.md` is updated to give a description of the API endpoint and sample request and response payloads?

kdamas93: @jamesaimonetti  Pr updated and ci passed

doc/getstream.md
```diff
@@ -8,10 +8,11 @@
 
 ## Create
 
-> PUT /v2/accounts/{ACCOUNT_ID}/getstream/token
+> PUT /v2/accounts/{ACCOUNT_ID}/getstream/connect
```
jamesaimonetti, line 11: It would be instructive to see what sample request JSON (if any) and sample response JSON users of this API should expect.

jamesaimonetti: It probably also warrants adding a section below for system administrators with instructions on how to setup the server-side stuff - what system_config doc(s) to configure, for instance. How does one go from base cluster to one with getstream integration.

kdamas93: > It would be instructive to see what sample request JSON (if any) and sample response JSON users of this API should expect.

Json response added :)

kdamas93: > It probably also warrants adding a section below for system administrators with instructions on how to setup the server-side stuff - what system_config doc(s) to configure, for instance. How does one go from base cluster to one with getstream integration.

Configuration notes added :)

PR #238, opened 2022-01-25
Fixed a bug in jonny5 to  allow the call limits to be set back to -1 through the api.

bradfordben:


lazedo:
@bradfordben if there's a bug in jonny5, then maybe fix it there instead of accepting the bug ?

bradfordben:
@lazedo Its not a bug in j5, its a bug in the API, -1 is a valid number for each of the fields but the only way you can set then to -1 is by editing the doc in couch as the schema is incorrect and CB just rejects the request during schema validation.

PR #241, opened 2022-02-05
KCAL-33: Include branch_extension documention for menus

harenson:
Related PR: https://github.com/2600hz/kazoo-callflow/pull/103
Depends on: https://github.com/2600hz/kazoo-core/pull/373


priv/api/swagger.json
```diff
@@ -8377,7 +8377,7 @@
                     "type": "array"
                 },
                 "hunt": {
-                    "default": true,
+                    "default": false,
```
lazedo: why is this changing ?

lazedo (changes requested):

priv/api/swagger.json
harenson: Because when you create a new menu, that value is false by default.

lazedo: and when did that changed without updating this ? afaics, its not related to this pr

harenson: I just reverted that change.

lazedo:
@harenson i think that `stop_after_extension` is not correct as the executing may not call `cf_exe:stop/1`, we should use `branch_extension` (or some rewording around that) that would reflect what we really do, ie, **branch** or **continue**

harenson:
@lazedo Yes, makes sense.

PR #242, opened 2022-02-10
KCRO-128: add paths to cowboy router on startup

icehess:
Allow other apps to configure a callback modules so when Crossbar is starting up
can call the handlers to get any custom routes to add to cowboy router.

The handler should implement `route_paths/0` function and return a list of valid
cowboy `cowboy_router:route_path()`.


src/crossbar_init.erl
```diff
@@ -44,6 +52,33 @@ deps_bytes() ->
...
+try_get_custom_path(CustomHandlerMod, Acc) ->
+    case kz_module:is_exported(CustomHandlerMod, 'route_paths', 0) of
+        'true' ->
+            do_try_get_custom_path(kz_term:to_atom(CustomHandlerMod), Acc);
+        'false' ->
+            Acc
+    end.
+
+-spec do_try_get_custom_path(kz_term:ne_binary(), [route_path()]) -> [route_path()].
+do_try_get_custom_path(CustomHandlerMod, Acc) ->
```
jamesaimonetti, line 72: Fix spec, `CustomHandlerMod` is `atom()`

jamesaimonetti: or `module()` to be helpful

PR #246, opened 2022-02-18
Play Wordle via API

jamesaimonetti:
Because why not? :)


jamesaimonetti:
Play from the shell using properly: https://github.com/2600hz/kazoo-properly/pull/70

PR #249, opened 2022-03-11
KZOO-222: OTP MFA provider

jamesaimonetti:
This PR allows KAZOO to act as an MFA provider for OTPs (one time
password). When configured properly, a user can now generate a QR Code
to scan into their authenticator application.

When authenticating, the first request will be challenged with a 401;
including "multi_factor_response" in the next request with the TOTP
code will then allow the user to authenticate (if the code matches
what KAZOO would expect).


jamesaimonetti:
core PR: https://github.com/2600hz/kazoo-core/pull/386

jamesaimonetti:
Testing here: https://github.com/2600hz/kazoo-properly/pull/71

PR #258, opened 2022-04-05
KZOO-172: Highlight meaning of call recording direction

harenson:
Related to https://github.com/2600hz/kazoo-callflow/pull/92


doc/accounts.md
```diff
@@ -781,3 +781,15 @@ curl -v -X POST \
     "status": "success"
 }
 ```
+
+### account call_recording
+
+Account call recording configuration is relative to FreeSwitch, which means, `inbound/outbound` depends on FreeSwitch standing point of view.
```
jamesaimonetti: `FreeSWITCH` when you reference the project name

PR #259, opened 2022-04-06
CP-42: allow impersonation using API auth token

jamesaimonetti:
Prior to this change, only auth tokens generated by user_auth for
superduper admins were permitted to impersonate another account's
user.

This change allows superduper auth tokens generated by API key as
well. This supports scripts that might generate the top-level auth
token via API key and generate an impersonation auth token for use by
the UI.


jamesaimonetti:
Test in https://github.com/2600hz/kazoo-properly/pull/74

PR #261, opened 2022-04-08
AXS-14: add api to configure kazoo_appex

icehess:
Adding some APIs to configure and start using Marketplace though UI.


icehess:
Required PR:
* https://github.com/2600hz/kazoo-core/pull/382

PR #264, opened 2022-04-12
KREG-3: Add optional from softphone (TeamMate) to MS Teams migration mechanism

harenson:
Dependency for https://github.com/2600hz/kazoo-registrar/pull/16


priv/api/swagger.json
```diff
@@ -6983,6 +6983,11 @@
                         }
                     },
                     "type": "object"
+                },
+                "webrtc": {
+                    "default": false,
```
lazedo: remove the `default`

lazedo (changes requested):
`webrtc` is determined by registration. device creds can be used from a webrtc or softphone or deskphone device. by putting a `default` you are forcing a value when the device is created or modified.

icehess (approved):

PR #266, opened 2022-04-14
CP-33: Update resources schema to validate server/realm

jamesaimonetti:
Prior to this change, if "server" or "realm" had leading whitespace,
commands to FreeSWITCH for origination would fail.

Now, check that the value is a proper IP address (subject to KAZOO
restrictions) or hostname (again subject to KAZOO restrictions).


jamesaimonetti:
Requires https://github.com/2600hz/kazoo-core/pull/399
Tests in https://github.com/2600hz/kazoo-properly/pull/75

PR #268, opened 2022-04-18
KZOO-229: add schema for appex client config

icehess:


icehess:
Circular dependent PR:
* https://github.com/2600hz/kazoo-core/pull/382

PR #271, opened 2022-04-23
[master] KCRO-134: handle empty modb/yodb/other_db list and return empty result set

icehess:
kazoo_modb/kazoo_yodb get_range used to be check for db existence when creating
the db range. This would potentially results in an empty list.

When `databases` option is an empty list, then `crossbar_view` uses account db. This
will break query if the request was actually meant for MODB/YODB dbs.

Instead we gracefully handle the empty list and set the db_name in Context to
undefined so we return empty result set.

Also this changes the behaviour of case when there is an empty db list and no db set in
Context. Before this we used to return error with `invalid_db_name`, but now we
just return a successful empty result.

Pro of this approach is that the client won't be alarmed with unnecessary error
result and just sees an empty result.

Cons of this is that we lose visibility of underlying issue. But I can argue that
this actually beneficial since it will force all CB developers to carefully check
for their options and set appropriate databases.


icehess:
## master PRs
* actual fix:
    *https://github.com/2600hz/kazoo-crossbar/pull/271
* related prs (not required to fix the original issue, just getting code up to standard):
*    https://github.com/2600hz/kazoo-cdr/pull/17
        https://github.com/2600hz/kazoo-cdr-push/pull/71

## 5.1 PRs
* actual fix:
    * https://github.com/2600hz/kazoo-crossbar/pull/288
* related prs (not required to fix the original issue, just getting code up to standard):
    * https://github.com/2600hz/kazoo-cdr/pull/20
    * https://github.com/2600hz/kazoo-cdr-push/pull/74

## PRs for 5.0
* actual fix:
    * https://github.com/2600hz/kazoo-crossbar/pull/272
* related prs (not required to fix the original issue, just getting code up to standard):
    * https://github.com/2600hz/kazoo-cdr/pull/18
    * https://github.com/2600hz/kazoo-cdr-push/pull/72


## Enhance get range modb and deal with db error result

* 5.0: https://github.com/2600hz/kazoo-core/pull/404
* 5.1: https://github.com/2600hz/kazoo-core/pull/441
* master: https://github.com/2600hz/kazoo-core/pull/403

PR #275, opened 2022-04-30
schema update for fix appex timeout

icehess:


icehess:
Circular PR: https://github.com/2600hz/kazoo-core/pull/411

PR #280, opened 2022-05-11
Add a minLength check to URL/URI schema properties

jamesaimonetti:


jamesaimonetti:
Includes https://github.com/2600hz/kazoo-core/pull/428

PR #281, opened 2022-05-11
[5.1] AXS-14: add schemas and api to configure kazoo_appex

icehess:


icehess:
Backport of #261, #268 and #275

Circular PR: https://github.com/2600hz/kazoo-core/pull/429

PR #282, opened 2022-05-13
restart kazoo_appex when enabling or linking

icehess:


src/modules/cb_apps_store.erl
```diff
@@ -400,30 +400,60 @@ patch(Context, ?MARKETPLACE) ->
...
+%%      * if node was up and we went from diable->enable
+%%      * if node was up and we went from enabled+unlinked->enabled+linked
+%% * don't do if:
+%%      * if node was up we disable/enable
+%%      * if node was up and we went from enabled+linked->disable+unlink->enabled+linked
+%% FIXME: this only restart in the current node, we need to restart across all node?!
+%% Or at least say this user somehow
+maybe_restart_kazoo_appex() ->
+    Restart = fun() ->
+                      _ = application:stop('kazoo_appex'),
```
jamesaimonetti, line 446: Does this need to run zone or cluster-wide?

icehess: well that is what i said in comment (well at least in my own word lol).
Do you want to do it in this PR?

icehess: Just didn't want to mess with since now.

jamesaimonetti: @icehess not necessarily, was just curious what happens when its running elsewhere too.

PR #283, opened 2022-05-14
[5.1] restart kazoo_appex when enabling or linking

icehess:
* restart appex on enable, link

* add a fixme comment about the danger of restarting kazoo_appex

* add another fix me comment


icehess:
backportof (#282)

PR #293, opened 2022-06-08
[5.1] Set switch nodename for existing call id

jamesaimonetti:
Switch-Nodename is used to select the routing key for publishing the
originate request. If Existing-Call-ID is present, find the media
server it is on so the right routing key is created to send the
originate request to the appropriate zone.


harenson (approved):
:+1:

PR #294, opened 2022-06-16
KZOO-248: followup for more read_only moves

jamesaimonetti:


jamesaimonetti:
Needs https://github.com/2600hz/kazoo-core/pull/447
Test updates in https://github.com/2600hz/kazoo-properly/pull/83

PR #297, opened 2022-06-22
[master] SHIP-42: Commland compatibility api added for version compatible client selection

swysor:


src/modules/cb_commland.erl
```diff
@@ -0,0 +1,182 @@
...
+-spec authorize(cb_context:context()) ->
+          boolean() |
+          {'stop', cb_context:context()}.
+authorize(Context) ->
+    authorize_commland(Context, cb_context:req_verb(Context), cb_context:req_nouns(Context)).
+
+-spec authorize(cb_context:context(), path_token()) ->
+          boolean() |
+          {'stop', cb_context:context()}.
+authorize(Context, _) ->
```
jamesaimonetti: It seems only `/commland/compatibility` is the supported path so I would make this clause explicit: `authorize(Context, ?COMPAT) ->` and remove the `authorize/3` function. Similar change for authenticate

src/modules/cb_commland.erl
```diff
@@ -0,0 +1,182 @@
...
+-spec authorize(cb_context:context(), path_token(), path_token()) ->
+          boolean() |
+          {'stop', cb_context:context()}.
+authorize(Context, _, _) ->
+    authorize_commland(Context, cb_context:req_verb(Context), cb_context:req_nouns(Context)).
+
+-spec authorize_commland(cb_context:context(), http_method(), req_nouns()) ->
+          boolean() |
+          {'stop', cb_context:context()}.
+authorize_commland(_, ?HTTP_GET, ?COMPAT_NOUNS) ->
```
jamesaimonetti: In general, we try not to use `_` in function heads, just prefix the variable name with `_` (like `_Context` in this case).

src/modules/cb_commland.erl
```diff
@@ -0,0 +1,182 @@
...
+authenticate_commland(_Context, _, _Nouns) ->
+    'false'.
+
+%%------------------------------------------------------------------------------
+%% @doc This function determines if the provided list of Nouns are valid.
+%% Failure here returns `404 Not Found'.
+%% @end
+%%------------------------------------------------------------------------------
+-spec resource_exists() -> 'true'.
+resource_exists() -> 'true'.
```
jamesaimonetti: Similar here; the only really valid path appears to be `/commland/compatibility` so remove functions that should never be called.

src/modules/cb_commland.erl
```diff
@@ -0,0 +1,182 @@
...
+%% @end
+%%------------------------------------------------------------------------------
+
+-spec validate(cb_context:context()) -> cb_context:context().
+validate(Context) ->
+    validate_commland(cb_context:set_db_name(Context, ?KZ_WEBHOOKS_DB), cb_context:req_verb(Context)).
+
+-spec validate(cb_context:context(), path_token()) -> cb_context:context().
+validate(Context, ?COMPAT) ->
+    validate_commland(cb_context:set_db_name(Context, ?KZ_WEBHOOKS_DB), cb_context:req_verb(Context)).
```
jamesaimonetti: Why setting the db name to webhooks?

jamesaimonetti (changes requested):

src/modules/cb_commland.erl
swysor: d'oh

src/modules/cb_commland.erl
swysor: gotcha, thanks

src/modules/cb_commland.erl
swysor: Do i need to worry about the path tokens here or is it ok for those to be _?

src/modules/cb_commland.erl
swysor: Ok, makes sense. I wasn't quite clear if i have to implement the /1 version of these as well. do  i only need the arity 2 for these?

swysor: removed those as well

priv/couchdb/schemas/system_config.crossbar.commland.json
```diff
@@ -0,0 +1,23 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "_id": "system_config.crossbar.commland",
+    "description": "Schema for crossbar.commland system_config",
+    "properties": {
+        "base_salt": {
+            "default": "7f613340-1ebf-4710-ad57-fb4d3041af4c",
```
jamesaimonetti, line 7: Do not set a default here. The `kapps_config:get` can set a default generated per cluster but it should definitely not be defined here.

priv/oas3/oas3-schemas.yml
```diff
@@ -10827,6 +10839,22 @@
       'description': crossbar.channels system wide channels list
       'type': boolean
   'type': object
+'system_config.crossbar.commland':
```
jamesaimonetti, line 10842: what's the diff between crossbar.commland and commland system_config docs?

src/modules/cb_commland.erl
```diff
@@ -0,0 +1,159 @@
...
+-include("crossbar.hrl").
+
+
+-define(DOT, <<".">>).
+-define(COMPAT, <<"compatibility">>).
+-define(COMPAT_NOUNS, [{<<"commland">>,[<<"compatibility">>]}]).
+-define(COMMLAND_CONFIG_CAT, <<"crossbar.commland">>).
+-define(BASE_URL, kapps_config:get_ne_binary(?COMMLAND_CONFIG_CAT, <<"base_url">>, <<"https://packages.2600hz.com/commland/misc/compatibility">>)).
+-define(BASE_VERSION, kapps_config:get_ne_binary(?COMMLAND_CONFIG_CAT, <<"base_version">>, <<"5.1">>)).
+-define(BASE_SALT, kapps_config:get_ne_binary(?COMMLAND_CONFIG_CAT, <<"base_salt">>, <<"7f613340-1ebf-4710-ad57-fb4d3041af4c">>)).
```
jamesaimonetti, line 30: Should look into a couple things:
1. generating a salt instead of hard-coding
2. can `kz_binary:rand_hex(16)` be enough?
3. some SUP methods in crossbar_maintenance to change the base salt

src/modules/cb_commland.erl
```diff
@@ -0,0 +1,159 @@
...
+    lager:debug("no compatible version found using base ~s", [?BASE_VERSION]),
+    to_hash(?BASE_VERSION).
+
+-spec to_dotted_version(kz_term:ne_binary(), kz_term:ne_binary()) -> kz_term:ne_binary().
+to_dotted_version(Major, Minor) ->
+    kz_binary:join([Major, Minor], ?DOT).
+
+-spec to_hash(kz_term:ne_binary()) -> kz_term:ne_binary().
+to_hash(Version) ->
+    kz_binary:md5(<<Version/binary, (?BASE_SALT)/binary>>).
```
jamesaimonetti: I think you can just do `kz_binary:md5([Version, ?BASE_SALT])` and it should give you an MD5 hash

jamesaimonetti (changes requested):

src/modules/cb_commland.erl
swysor: it has to be deterministic and is meant to obscure the version of the core it is querying. the core produces a hash that is also produced on the shipyard server serving  the packages. The URL here is the compatible path to upgrade commland.

src/modules/cb_commland.erl
swysor: righto, ill fix.

priv/oas3/oas3-schemas.yml
swysor: oh i realized that i should be making a crossbar.commland not a commland system config doc since its parameters are only related to this api and i didn't want to step on future development with kazoo and commland. I can change it back to just commland if you would prefer.

PR #305, opened 2022-06-27
CP-47: cascading account changes

jamesaimonetti:
Allow admins to disable an account and its sub-accounts. Additionally,
allow admins to disable based on the account's balance when below a
threshold.


jamesaimonetti:
Testing in https://github.com/2600hz/kazoo-properly/pull/84

PR #325, opened 2022-07-28
Intra account impersonation enable/disable functionality added

nkarunarathne:


src/modules/cb_user_auth.erl
```diff
@@ -139,9 +139,22 @@ authorize_nouns(_, _Nouns, _) -> 'false'.
...
-    lager:info("auth token's account id matches request's account id ~s, not allowed", [RequestAccountId]),
-    forbidden(Context);
+maybe_authorize_impersonation(Context, RequestAccountId, RequestUserId, RequestAccountId) ->
+    {'ok', RootAccountId} = kapps_util:get_master_account_id(),
+    AuthToken = cb_context:auth_token(Context),
+    {'ok', AuthJObj} = kz_auth:validate_token(AuthToken),
+    Method = kz_json:get_string_value(<<"method">>, AuthJObj),
+    case
+        (RootAccountId == RequestAccountId)
+        andalso (Method == "cb_api_auth")
```
jamesaimonetti, line 149: Let's prefer binaries over strings unless required to use strings. `Method = kz_json:get_ne_binary_value(<<"method">>, AuthJObj)` should be preferred, as binary comparison is more efficient that list comparison.

jamesaimonetti: Also, it is preferred to use `=:=` and `=/=` for equality checks.

src/modules/cb_user_auth.erl
```diff
@@ -139,9 +139,22 @@ authorize_nouns(_, _Nouns, _) -> 'false'.
...
+    of
+        false ->
```
jamesaimonetti, line 151: `'false'` and `'true'` - ticky the atoms :) you can run `make elvis` from the kazoo root dir to help highlight these.

jamesaimonetti (changes requested):

src/modules/cb_user_auth.erl
```diff
@@ -139,9 +139,22 @@ authorize_nouns(_, _Nouns, _) -> 'false'.
 maybe_authorize_impersonation(Context, _RequestAccountId, _RequestUserId, 'undefined') ->
     lager:info("no auth account on auth token"),
     forbidden(Context);
-maybe_authorize_impersonation(Context, RequestAccountId, _RequestUserId, RequestAccountId) ->
-    lager:info("auth token's account id matches request's account id ~s, not allowed", [RequestAccountId]),
-    forbidden(Context);
+maybe_authorize_impersonation(Context, RequestAccountId, RequestUserId, RequestAccountId) ->
```
icehess, line 142: If the the intend is to allow only `cb_api_auth` when the account ids are same, then change it to something like this:

```
    lager:info("auth token's account id matches request's account id ~s, checking if auth method was cb_api_token", [RequestAccountId]),
    case cb_context:is_superduper_admin(Context) of
        'false' ->
            lager:info("auth account id ~s is not superduper", [AuthAccountId]),
            forbidden(Context);
        'true' ->
            maybe_authorize_impersonation_via_api_token(Context, RequestAccountId, RequestUserId, AuthAccountId
                                                           ,kz_json:get_value(<<"method">>, cb_context:auth_doc(Context))
                                                           );
    end;
```

src/modules/cb_user_auth.erl
```diff
@@ -139,9 +139,22 @@ authorize_nouns(_, _Nouns, _) -> 'false'.
...
+    {'ok', RootAccountId} = kapps_util:get_master_account_id(),
+    AuthToken = cb_context:auth_token(Context),
+    {'ok', AuthJObj} = kz_auth:validate_token(AuthToken),
```
icehess, line 145: fyi We already validate the token! you just to `kz_json:get_string_value(<<"method">>, cb_context:auth_doc(Context))`,

But anyway remove this and use the code from comment above.

icehess (changes requested):

PR #334, opened 2022-08-17
coc-166-printingConferenceNumberInsteadId

kdamas93:
https://github.com/2600hz/kazoo-core/pull/479


src/modules/cb_conferences.erl
```diff
@@ -429,7 +429,7 @@ build_email_invite_kapi_req(Context) ->
                                               ),
     Conf = cb_context:fetch(Context, 'conference_doc'),
     ReqData = cb_context:req_data(Context),
-    [{<<"Conference-ID">>, kz_doc:id(Conf)}
+    [{<<"Conference-ID">>, kzd_conferences:conference_numbers(Conf)}
```
jamesaimonetti, line 432: Sorry, why would the conference ID be the list of conference numbers? I think the template being used should list the conference number(s) separately, or provide instructions to use the conference number to login instead of conference ID.

I don't think this requires a code fix so much as updating the template to be more clear on which value to use for joining the conference.

jamesaimonetti (changes requested):

kdamas93:
The changes were implemented in the following prs:
https://github.com/2600hz/kazoo-core/pull/490
https://github.com/2600hz/kazoo-crossbar/pull/337
https://github.com/2600hz/kazoo-teletype/pull/47

PR #337, opened 2022-09-05
coc-254-ReplacingConfenceNumberInsteadId

kdamas93:
https://github.com/2600hz/kazoo-core/pull/490
https://github.com/2600hz/kazoo-teletype/pull/47


src/modules/cb_conferences.erl
```diff
@@ -429,7 +429,7 @@ build_email_invite_kapi_req(Context) ->
                                               ),
     Conf = cb_context:fetch(Context, 'conference_doc'),
     ReqData = cb_context:req_data(Context),
-    [{<<"Conference-ID">>, kz_doc:id(Conf)}
```
jamesaimonetti, line 432: I don't think you need to remove conference-id, you can just add conference-numbers, right?

jamesaimonetti (changes requested):

src/modules/cb_conferences.erl
kdamas93: Changes made for master and 5.1 :) , thanks James !

icehess (approved):

PR #342, opened 2022-09-12
coc-254-Refactor-5.1

kdamas93:
https://github.com/2600hz/kazoo-teletype/pull/52


src/modules/cb_conferences.erl
```diff
@@ -434,13 +434,18 @@ build_email_invite_kapi_req(Context) ->
...
+    ,{<<"Conference-Call-In-Numbers">>, maybe_remove_not_numeric_values([kz_json:get_binary_value(<<"numbers">>, CF) || CF <- Callflows])}
     ,{<<"Conference-Participant-Pins">>, kzd_conferences:member_pins(Conf)}
     ,{<<"Invite-Message">>, kz_json:get_binary_value(<<"invite_message">>, ReqData)}
     ,{<<"Guests">>, kz_json:get_list_value(<<"guests">>, ReqData)}
     | kz_api:default_headers(?APP_NAME, ?APP_VERSION)
     ].
 
+-spec maybe_remove_not_numeric_values(list()) -> list().
+maybe_remove_not_numeric_values(ConferenceCallInNumbers)->
+    ToFilterOut = [<<"undefinedconf">>],
```
jamesaimonetti, line 446: If you only have the one value, hard-coded, you don't need to call lists:member below.
```erlang
[N || N <- ConferenceCallInNumbers, N =/= <<"undefinedconf">>].
```
Or `lists:delete(<<"undefinedconf">>, ConferenceCallInNumbers)` if only one instance of that value is present.

jamesaimonetti (changes requested):

src/modules/cb_conferences.erl
icehess: Also move this logic to teletype itself. less code change and prs to merge.

icehess: @jamesaimonetti is it safe to use `knm_converters:is_reconcilable/1` here? I don't think we ever use or even allow to use extension for conference?

jamesaimonetti: There's no restriction on using extensions for a conference that I'm aware of.

kdamas93:
Moving logic to teletype, less code change and prs to merge.

kdamas93:
Moving logic to teletype, less code change and prs to merge.

PR #354, opened 2022-10-04
PS-60: conference API to set vars

jamesaimonetti:


src/modules/cb_conferences.erl
```diff
@@ -461,9 +461,27 @@ handle_conference_action(Context, ConferenceId, <<"dial">>) ->
     dial(Context, ConferenceId, cb_context:req_value(Context, <<"data">>));
 handle_conference_action(Context, ConferenceId, <<"record">>) ->
     record_conference(Context, ConferenceId, cb_context:req_value(Context, <<"data">>));
+handle_conference_action(Context, ConferenceId, <<"vars">>) ->
+    vars_conference(Context, ConferenceId);
 handle_conference_action(Context, ConferenceId, Action) ->
     lager:error("unhandled conference id ~p action: ~p", [ConferenceId, Action]),
-    cb_context:add_system_error('faulty_request', Context).
+    Error = kz_json:from_list([{<<"message">>, <<"Value not found in enumerated list of values">>}
+                              ,{<<"target">>, [<<"lock">>, <<"unlock">>, ?PLAY, <<"dial">>, <<"record">>, <<"vars">>]}
```
harenson, line 469: Why not creating/updating a schema to list these values as valid for the `action` field, and then validate request with the schema before handling conference actions?

jamesaimonetti: Time, tbh, Just trying to get the feature delivered. Would be a good refactor for another day though

PR #355, opened 2022-10-06
kzoo-299-API_not_returning_DNS_settings-master

kdamas93:


src/modules/cb_whitelabel.erl
```diff
@@ -510,7 +510,7 @@ test_host(Host, HostConfig, DomainType, Options) ->
           kz_term:ne_binaries().
 lookup(Host, DomainType, Options) ->
     Type = kz_term:to_atom(kz_term:to_lower_binary(DomainType)),
-    {'ok', Lookup} = kz_network_utils:lookup_dns(Host, Type, Options),
+    {'ok', Lookup} = kz_network_utils:lookup_dns(string:trim(Host), Type, Options),
```
jamesaimonetti: `Host` is ne_binary(); check kz_binary for a trim function

PR #362, opened 2022-10-21
KZOO-335: add call limits schema

harenson:
Depends on: https://github.com/2600hz/kazoo-core/pull/512

Allows to set a maximum number of concurrent calls (limit) per endpoint (device/user) or account-wide. When building the endpoint, if limit is configured at endpoint level, it takes precedence over account's configuration in case limit was also configured account-wide.


danielfinke (approved):

priv/couchdb/schemas/call_limits.json
```diff
@@ -0,0 +1,12 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "_id": "call_limits",
+    "description": "Schema for call limits",
+    "properties": {
+        "max_concurrent": {
+            "description": "Maximum number of concurrent calls allowed per endpoint",
+            "type": "integer"
+        }
```
lazedo, line 9: lets add a `direction` property with enum inbound/outbound/both and default to `both`

lazedo (changes requested):

priv/couchdb/schemas/call_limits.json
harenson: @lazedo should we stick with the existing `[inbound, outbound, any]` values for direction?

lazedo: @harenson nm, lets leave it as-is

harenson: @lazedo Ok, I will take care of the comments on the other related PRs, then. Thank you.

harenson:
_docs_ check is failing because it depends on https://github.com/2600hz/kazoo-core/pull/512 changes.

lazedo (approved):

PR #368, opened 2022-10-31
KCRO-157 : Additional Restrictions on Voicemail Message API

sanjeevie6353:
KCRO-157 : Additional Restrictions on Voicemail Message API

Added system config flag that if enabled will cause the voicemail message API listing and retrieval to only succeed if the JWT belongs to an admin or if it belongs to a user that user is the owner of the voicemail box the message belongs to.


priv/couchdb/schemas/system_config.crossbar.json
```diff
@@ -3,6 +3,11 @@
     "_id": "system_config.crossbar",
     "description": "Schema for crossbar system_config",
     "properties": {
+        "additional_restriction_on_request": {
+            "default": false,
+            "description": "crossbar additional_restriction_on_request",
```
jamesaimonetti: you can populate this with a proper description, please

src/cb_context.erl
```diff
@@ -1357,3 +1359,11 @@ add_metadata_value(#cb_context{metadata=Metadata}=Context, Key, 'null') ->
 add_metadata_value(#cb_context{metadata=Metadata}=Context, Key, Value) ->
     lager:info("setting meta ~p: ~p", [Key, Value]),
     Context#cb_context{metadata = props:set_value(Key, Value, Metadata)}.
+
+%% @doc get owner id from auth doc (of cb_)
+-spec auth_doc_owner_id(context()) -> kz_term:api_ne_binary().
+auth_doc_owner_id(Context) ->
+    case kz_json:get_value(<<"owner_id">>, auth_doc(Context)) of
```
jamesaimonetti: i think auth_user_id already does similar, no?

src/modules/cb_simple_authz.erl
```diff
@@ -131,6 +136,75 @@ account_is_descendant('false', Context, AuthAccountId) ->
...
     end.
 
+%%------------------------------------------------------------------------------
+%% @doc Returns true if additional authorizationcheck is not needed at
+%% cb_token_restrictions module
+%% @end
+%%------------------------------------------------------------------------------
+-spec no_additional_restrictions() -> boolean().
+no_additional_restrictions() ->
+    case kapps_config:get_is_true(?CONFIG_CAT, <<"additional_restriction_on_request">>, 'false') of
```
jamesaimonetti: You define a macro then don't use it?

src/modules/cb_simple_authz.erl
```diff
@@ -131,6 +136,75 @@ account_is_descendant('false', Context, AuthAccountId) ->
...
+        'false' -> 'true';
```
jamesaimonetti: may as well just `not kapps_config:is_true(...)` vs a case, right?

src/modules/cb_simple_authz.erl
```diff
@@ -131,6 +136,75 @@ account_is_descendant('false', Context, AuthAccountId) ->
...
+
+%%------------------------------------------------------------------------------
+%% @doc Returns true if the requested account id is an superduper admin, admin
+%% or the user is the owner of the requested data
+%% @end
+%%------------------------------------------------------------------------------
+-spec request_allowed_for_user(cb_context:context()) -> boolean().
+request_allowed_for_user(Context) ->
+    AuthAccountId = cb_context:auth_account_id(Context),
+    request_allowed_for_user(cb_context:is_superduper_admin(AuthAccountId), Context).
```
jamesaimonetti: Should just pass `cb_context:is_superduper_admin(Context)`; why use AuthAccountId instead?

src/modules/cb_simple_authz.erl
```diff
@@ -131,6 +136,75 @@ account_is_descendant('false', Context, AuthAccountId) ->
...
+                'true' ->
+                    validate_request_for_user(Context, cb_context:path_tokens(Context))
+            end
+    end.
+
+%%------------------------------------------------------------------------------
+%% @doc Returns true if user is the owner of the requested data
+%% @end
+%%------------------------------------------------------------------------------
+-spec validate_request_for_user(cb_context:context(), kz_term:ne_binary()) -> boolean().
```
jamesaimonetti: This should probably not be defined here but in cb_vmboxes, in the authorize/N callbacks

jamesaimonetti (changes requested):

priv/couchdb/schemas/system_config.crossbar.json
sanjeevie6353: Changed the description of config parameter to **"crossbar allow request for only admins and owner of data"** on the following.
- swagger.json
- system_config.crossbar.json
- oas3-schemas.yml

Currently this is considered as a common config value. The aim is to be used for any other user level restrictions along with restrictions on VMBox message. Should this config only be related to VMBox messages?

src/cb_context.erl
sanjeevie6353: Yes it seems to do the same. Changed the logic to use the auth_user_id() function and removed auth_doc_owner_id()

src/modules/cb_simple_authz.erl
sanjeevie6353: Agreed on the change. Altered function to use not kapps_config:is_true(...) and removed the no_additional_restrictions() function

src/modules/cb_simple_authz.erl
sanjeevie6353: This has been removed.

src/modules/cb_simple_authz.erl
sanjeevie6353: Used an already available function at  `**authorize(Context, Verb, _Nouns)**`  function (in cb_simple_authz.erl). I've changed my function to use `cb_context:is_superduper_admin(Context)`.

Shall I change other places that use AuthAccountId?

src/modules/cb_simple_authz.erl
sanjeevie6353: The intention of defining these functions in the cb_simple_authz.erl was to allow new api urls to be defined in new clauses. 

This will allow to restrict url access on requests other than vm_box_messages. 

If I'm to put this inside cb_vmboxes.erl, it will restrict this functionality only to vmboxes. So, Is it ok to keep the current implementation intact?

jamesaimonetti: no, as we have an `authorize` available for each endpoint that needs to define custom rules. Let's keep the `simple` in `cb_simple_authz`.

src/modules/cb_simple_authz.erl
jamesaimonetti: no, only places you've touched. Will have to evaluate other instances to find out why they are as they are. Thanks for looking into it

src/modules/cb_simple_authz.erl
sanjeevie6353: Did the change to move functionality to cb_vmboxes

src/modules/cb_token_restrictions.erl
```diff
@@ -138,6 +138,15 @@ delete(Context) ->
           'false' |
           {'true' | 'stop', cb_context:context()}.
 authorize(Context) ->
+    case props:get_value(<<"vmboxes">>, cb_context:req_nouns(Context)) of
```
jamesaimonetti: This change is unnecessary. cb_vmboxes:authorize should return true/false and things should just work. This module has nothing to do with vmboxes and should have no logic associated with it. I think the vmboxes module's change should be sufficient for whether to allow the request or not.

src/modules/cb_vmboxes.erl
```diff
@@ -140,6 +140,63 @@ allowed_methods(_VMBoxId, ?MESSAGES_RESOURCE, _VMMsgId) ->
...
 
+
+%%------------------------------------------------------------------------------
+%% @doc Returns true if the requested account id is an superduper admin, admin
+%% or the user is the owner of the requested data
+%% @end
+%%------------------------------------------------------------------------------
+-spec authorize(cb_context:context()) -> boolean().
+authorize(Context) ->
+    authorize(cb_context:is_superduper_admin(Context), Context).
```
jamesaimonetti: As a general request, keep argument order when calling sub-functions. There's a minor speed improvement with maintaining order compared to moving their positions around.
So `authorize(Context, cb_context:is_superduper_admin(Context)).`

jamesaimonetti: And it just bugs me lol

src/modules/cb_vmboxes.erl
```diff
@@ -140,6 +140,63 @@ allowed_methods(_VMBoxId, ?MESSAGES_RESOURCE, _VMMsgId) ->
...
+%% @end
+%%------------------------------------------------------------------------------
+-spec validate_request_for_user(cb_context:context(), kz_term:ne_binary()) -> boolean().
+validate_request_for_user(Context, [_,_,<<"accounts">>,_,<<"vmboxes">>,VMBoxID,<<"messages">>,_]) ->
+    VMBox = crossbar_doc:load(VMBoxID, Context, ?TYPE_CHECK_OPTION(kzd_vmboxes:type())),
+    OwnerId = cb_context:auth_user_id(Context),
+    case kz_json:get_value(<<"owner_id">>, cb_context:doc(VMBox)) of
+        OwnerId -> 'true';
+        'undefined' ->
+            lager:debug("validate request for user failed for Owner: ~p", [OwnerId]),
```
jamesaimonetti: I would say something more like "vmbox ~s has no owner, auth user ~s cannot access this box", [VMBoxId, AuthUserId]

src/modules/cb_vmboxes.erl
```diff
@@ -140,6 +140,63 @@ allowed_methods(_VMBoxId, ?MESSAGES_RESOURCE, _VMMsgId) ->
...
+    end.
+
+%%------------------------------------------------------------------------------
+%% @doc Returns true if user is the owner of the requested data
+%% @end
+%%------------------------------------------------------------------------------
+-spec validate_request_for_user(cb_context:context(), kz_term:ne_binary()) -> boolean().
+validate_request_for_user(Context, [_,_,<<"accounts">>,_,<<"vmboxes">>,VMBoxID,<<"messages">>,_]) ->
+    VMBox = crossbar_doc:load(VMBoxID, Context, ?TYPE_CHECK_OPTION(kzd_vmboxes:type())),
+    OwnerId = cb_context:auth_user_id(Context),
```
jamesaimonetti: This should be `AuthUserId`

src/modules/cb_vmboxes.erl
```diff
@@ -140,6 +140,63 @@ allowed_methods(_VMBoxId, ?MESSAGES_RESOURCE, _VMMsgId) ->
...
+    case kz_json:get_value(<<"owner_id">>, cb_context:doc(VMBox)) of
+        OwnerId -> 'true';
```
jamesaimonetti: This should be `AuthUserId` as well

src/modules/cb_vmboxes.erl
```diff
@@ -140,6 +140,63 @@ allowed_methods(_VMBoxId, ?MESSAGES_RESOURCE, _VMMsgId) ->
...
+        'undefined' ->
+            lager:debug("validate request for user failed for Owner: ~p", [OwnerId]),
+            'false';
+        _OtherOwnerId  ->
+            lager:debug("validate request for user failed - Requester is not the Owner: ~p", [_OtherOwnerId]),
```
jamesaimonetti: "access to vmbox ~s for auth user ~s denied, box is owned by ~s", [VMBoxId, AuthUserId, _OwnerId]

src/modules/cb_vmboxes.erl
```diff
@@ -140,6 +140,63 @@ allowed_methods(_VMBoxId, ?MESSAGES_RESOURCE, _VMMsgId) ->
...
+-spec authorize(boolean(), cb_context:context()) -> boolean().
+authorize('true', _Context) -> 'true';
+authorize('false', Context) ->
+    case cb_context:is_account_admin(Context) of
+        'true' -> 'true';
+        'false' ->
+            case kapps_config:get_is_true(?CONFIG_CAT, <<"restriction_on_vmbox_api">>, 'false') of
+                'false' -> 'true';
+                'true' ->
+                    validate_request_for_user(Context, cb_context:path_tokens(Context))
```
jamesaimonetti: Use `cb_context:req_nouns(Context)` instead, you'll get the list of endpoints (in reverse order) plus their arguments.

`validate_request_for_user` then becomes:
```
validate_request_for_user(Context, [{<<"vmboxes">>, [VMBoxId | _ ]}, {<<"accounts">>, [AccountId]}]) ->
    ...
```
Basically any API under vmboxes, with a Box ID, should be checked for ownership (so /vmboxes/ID, /vmboxes/ID/messages and /vmboxes/ID/messages/MSG_ID)

jamesaimonetti (changes requested):

src/modules/cb_vmboxes.erl
sanjeevie6353: I did not know that tbh... I'll take a note on this... Thank you...

This change has been done..

src/modules/cb_vmboxes.erl
sanjeevie6353: Changed this to `AuthUserId`

src/modules/cb_vmboxes.erl
sanjeevie6353: Changed this to AuthUserId

src/modules/cb_vmboxes.erl
sanjeevie6353: Changed the log

src/modules/cb_vmboxes.erl
sanjeevie6353: Changed the log

src/modules/cb_token_restrictions.erl
sanjeevie6353: This logic has been removed.

src/modules/cb_vmboxes.erl
sanjeevie6353: This has been fixed.

jamesaimonetti:
Tests in https://github.com/2600hz/kazoo-properly/pull/103

PR #372, opened 2022-11-09
[master] KCRO-162: Improve security of the monster-ui forgot password feature

filipevarjao:
A vulnerability on the reset password endpoint allowing identify valid accounts by using the reset password requests, so now we always return success when requested: An email will be sent with details on how to reset your password.
In addition, this request consumes tokens from token buckets.

Tests https://github.com/2600hz/kazoo-properly/pull/139


src/modules/cb_user_auth.erl
```diff
@@ -595,9 +602,12 @@ maybe_load_user_doc_by_username(Account, Context) ->
             cb_context:add_validation_error(<<"username">>, <<"not_found">>, Msg, Context)
     end.
 
--spec save_reset_id_then_send_email(cb_context:context()) -> cb_context:context().
-save_reset_id_then_send_email(Context) ->
-    MoDb = kazoo_modb:get_modb(cb_context:db_name(Context)),
+-spec save_reset_id_then_send_email(cb_context:context(), kz_term:ne_binary()) -> cb_context:context().
+save_reset_id_then_send_email(Context, 'undefined') ->
```
jamesaimonetti, line 595: How can the MODB ever be 'undefined'? get_modb above would crash first, no?

filipevarjao: it will be 'undefined' when the data provided is invalid, which is the issue here.
cb_context:db_name(Context) with an invalid username/email/account name/phone number returns undefined
then kazoo_modb:get_modb('undefined') will also return 'undefined'

src/modules/cb_user_auth.erl
```diff
@@ -760,41 +769,23 @@ find_account([{'id', AccountId} | Types], Context) ->
...
 -spec error_no_account_phone_number(cb_context:context(), kz_term:ne_binary()) -> cb_context:context().
 error_no_account_phone_number(Context, PhoneNumber) ->
-    Msg =
-        kz_json:from_list(
-          [{<<"message">>, <<"The provided phone number could not be found">>}
-          ,{<<"cause">>, PhoneNumber}
-          ]),
-    cb_context:add_validation_error(<<"phone_number">>, <<"not_found">>, Msg, Context).
+    lager:debug("failed to find phone number: '~s'", [PhoneNumber]),
+    Context.
```
icehess: This function also can be called when you're authenticating (via `find_account`, `validate_action/2` and `validate/1`), no? Not returning error will allow calls to `crossbar_auth:create_auth_token/2` on `put/1` when a user is trying to login.

Find another way to return success if it is a recovery request.

icehess (changes requested):

src/modules/cb_user_auth.erl
```diff
@@ -256,6 +256,12 @@ validate_action(Context, _Action) ->
 
 -spec validate(cb_context:context(), path_token()) -> cb_context:context().
 validate(Context, ?RECOVERY) ->
+    Context1 = consume_tokens(Context),
```
jamesaimonetti: Check the return of `consume_tokens/1` here

src/modules/cb_user_auth.erl
```diff
@@ -591,9 +599,12 @@ maybe_load_user_doc_by_username(Account, Context) ->
             cb_context:add_validation_error(<<"username">>, <<"not_found">>, Msg, Context)
     end.
 
--spec save_reset_id_then_send_email(cb_context:context()) -> cb_context:context().
-save_reset_id_then_send_email(Context) ->
-    MoDb = kazoo_modb:get_modb(cb_context:db_name(Context)),
+-spec save_reset_id_then_send_email(cb_context:context(), kz_term:ne_binary()) -> cb_context:context().
```
jamesaimonetti: Compare the spec to the args in the function clauses

src/modules/cb_user_auth.erl
```diff
@@ -724,7 +733,8 @@ find_account([{'name', AccountName} | Types], Context) ->
             lager:debug("the account name returned multiple results"),
             {'ok', AccountDbs};
         {'error', _} ->
-            find_account(Types, error_no_account_name(Context, AccountName))
+            Paths = cb_context:path_tokens(Context),
```
icehess: Instead of changing the `find_account/2` directly, you can handle its result right in `maybe_load_user_doc_via_creds/1`. 

Also I suggest renaming `maybe_load_user_doc_via_creds` to `find_user_for_password_recovery` as it is more clear what is this function for.

If `find_account/1` returns error tuple, then do this in `maybe_load_user_doc_via_creds/1` case:

```erlang
  -spec find_user_for_password_recovery(cb_context:context()) -> cb_context:context().
  find_user_for_password_recovery(Context) ->
      JObj = cb_context:doc(Context),
      Options = [{'phone_number', kzd_user_auth:phone_number(JObj)}
                ,{'realm', kz_json:get_first_defined([<<"account_realm">>, <<"realm">>], JObj)}
                ,{'name', kzd_accounts:normalize_name(kz_json:get_value(<<"account_name">>, JObj))}
                ,{'id', kz_json:get_ne_binary_value(<<"account_id">>, JObj)}
                ],
      case find_account(Options, Context) of
          {'error', _Context1} ->
              lager:debug("failed to find account, responding empty result for password recovery"),
              crossbar_doc:handle_json_success(kz_json:new(), Context);
          {ok, [Account|_]} ->
        ....
      end.
```

And also update the `maybe_load_user_doc_by_username/2` to return success instead of not_found:
```erlang
maybe_load_user_doc_by_username(Account, Context) ->
...
          _ ->
              Reason = kz_term:to_binary(io_lib:format("The provided user name ~s was not found", [Username])),
              crossbar_auth:log_failed_auth(?MODULE, AuthType, Reason, Context, Account),
              crossbar_doc:handle_json_success(kz_json:new(), Context)
    end.
```

This is less code, more explicit, let `find_account/1` do it jobs and it completely separates the logic of password reset from the functionality of _finding the account_. 

It also avoids `maybe_load_user_doc_by_username` to return user not_found.

src/modules/cb_user_auth.erl
```diff
@@ -735,7 +745,8 @@ find_account([{'realm', AccountRealm} | Types], Context) ->
             lager:debug("the account realm returned multiple results"),
             {'ok', AccountDbs};
         {'error', _} ->
-            find_account(Types, error_no_account_realm(Context, AccountRealm))
+            Paths = cb_context:path_tokens(Context),
+            find_account(Types, error_no_account_realm(Context, lists:last(Paths), AccountRealm))
```
icehess: Revert all of these changes here and from clauses below and from `error_*` functions, handle the return of `find_account/1` in `maybe_load_user_doc_via_creds/1`. See the note above.

src/modules/cb_user_auth.erl
```diff
@@ -591,9 +598,12 @@ maybe_load_user_doc_by_username(Account, Context) ->
...
     end.
 
--spec save_reset_id_then_send_email(cb_context:context()) -> cb_context:context().
-save_reset_id_then_send_email(Context) ->
-    MoDb = kazoo_modb:get_modb(cb_context:db_name(Context)),
+-spec save_reset_id_then_send_email(cb_context:context(), kz_term:ne_binary() | 'undefined') -> cb_context:context().
+save_reset_id_then_send_email(Context, 'undefined') ->
+    lager:debug("it failed but we must validate it to avoid leaking information"),
+    Cntxt = cb_context:set_validation_errors(Context, kz_json:new()),
+    cb_context:set_resp_status(Cntxt, 'success');
```
icehess:
```suggestion
    lager:debug("failed to find user for password recovery, sending success result"),
    Context1 = cb_context:set_validation_errors(Context, kz_json:new()),
    cb_context:set_resp_status(Context1, 'success');
```

Also please change the `Msg` in the next function clause from `Msg = <<"Request for password reset handled, email sent to: ", Email/binary>>,` to:

```
Msg = <<"Password resend email was sent, please check your email to learn how to reset your password.">>,
```

This is also hides the fact we didn't find the user or account by responding same generic message as if _we find the user_.

icehess (changes requested):
Some comments about separating the password recovery logic from `find_account/1`. Also I asked to change the messaging regarding the password reset request to same generic so an attacker cannot use different messaging response to brute force as it was intended in the ticket itself.

Also I think we may need to check the account is enabled and not expired, either before generating a password reset id, or before creating an auth token via password recovery link. Neither of these actually check that.

So for example **after** `maybe_load_user_doc_by_username/2` **finds the user**, it sets its stuff in context, and calls `maybe_account_is_expired/2` which properly checks the account.

icehess (approved):

PR #373, opened 2022-11-09
[master] - Voicemail envelope prompts customization to enable customized prompt

sanjeevie6353:
KCAL-46 : Voicemail envelope prompts customization to enable customized prompt

Currently the voicemail envelope will only allow enabling and disabling the envelope to play or ignore the envelope part of the voicemail message announcement.

This new feature will allow to select a specific envelope prompt based on the need. It will be configurable per voicemailbox.


sanjeevie6353:
Related PRs

https://github.com/2600hz/kazoo-core/pull/523
https://github.com/2600hz/kazoo-callflow/pull/166
https://github.com/2600hz/kazoo-sounds/pull/28

priv/couchdb/schemas/vmboxes.json
```diff
@@ -30,6 +30,10 @@
             "support_level": "supported",
             "type": "boolean"
         },
+        "envelope_type": {
```
jamesaimonetti, line 33: How is a client supposed to know what value(s) are valid here?

I would suggest using `"enum":["default", "caller_and_time"]` so the possible values are known, with "default":"default" to fill in when folks re-save a vmbox without setting this property.

sanjeevie6353: I added this and it works well now... Thank you for pointing this out... 

image

PR #374, opened 2022-11-09
kcro-148-CommlandSelectorApi-master

kdamas93:
The purpose of this API is to replicate the behavior of the **apps_store** API with the following differences:

-It will be possible to perform actions on the following resources through the **cb_application** endpoint
-application document which includes screenshots, icons, entitlement, and blocklist among others.

The creation and management of permissions for this will be in each account, unlike the current scheme that was only allowed in the master account.

This API will also allow handling different types of applications, for example, **commland_ui** and **monster_ui** , a field has been added which specifies the type of application that has been created. With this it is hoped to be able to have controlled and grouped the different applications that may arise within their category.

It has created its own module in core to execute actions from the console, the module is called **kazoo_ui_apps_maintenance**, and the related pr is https://github.com/2600hz/kazoo-core/pull/524, it has been developed based on **kazoo_documents_maintenance**, keeping the functionality that exists for monster-ui applications that follow the current architecture

The API contains its own schema, views and functions


src/crossbar_maintenance.erl
```diff
@@ -950,3 +963,204 @@ db_init() ->
 register_views() ->
     ?SUP_LOG_WARNING("crossbar register_views()"),
     kz_datamgr:register_views_from_folder('crossbar').
+
+%%------------------------------------------------------------------------------
+%% @doc U.
```
jamesaimonetti: What is this comment for?

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1200 @@
+%%%-----------------------------------------------------------------------------
+%%% @copyright (C) 2011-2022, 2600Hz
+%%% @doc Endpoint for Applications
+%%% @author Kevin Damas
+%%% @end
+%%%-----------------------------------------------------------------------------
+-module(cb_applications).
+
+-export([init/0
+        , authenticate/1
```
jamesaimonetti: please follow project formatting standards and remove the space between the comma and function/arity

doc/ref/applications.md
```diff
@@ -0,0 +1,218 @@
+# Applications
```
jamesaimonetti, line 1: This ref doc needs to be migrated to `crossbar/doc/applications/md` and be filled in appropriately.

src/cb_apps_util.erl
```diff
@@ -345,6 +364,37 @@ load_default_apps(ReturnAsMap) ->
 
     end.
 
+-spec load_default_applications() -> kz_json:objects().
+load_default_applications() ->
+    {'ok', MasterAccountDb} = kapps_util:get_master_account_db(),
+    ViewOptions = [{'startkey', [<<"application">>]},
+                   {'endkey', [<<"application">>, kz_json:new()]},
+                   'include_docs'],
```
jamesaimonetti: drop end of list char `]` to newline
comma-first for the list elements in here:

```erlang
[{startkey, [..]}
 ,{endkey, [...]}
 ,include_docs
],
```

icehess:
@kdamas93 I don't understand the need for new functions. These functions should be same for both MUI, commland or any other app type in the future.

You can simply repurpose the current functions or enhance them to get what you want.

Also I don't see anything about `application_type` here.

Another thing, if you look `init_apps` is actually moved to `kazoo_document_maintenance`. This the place you have to write the `init_applications`. This function must be in core, since kazoo_appex also depends on it.

kdamas93:
@icehess 
What functions are you referring to?
I tried to adapt and reuse as much as possible, I would appreciate it if you could tell me what they are:

Regarding the logic related to `application_type`, it is contained in `cb_applications`:

I will start working on the observation about `init_apps`/`init_applications`

icehess:
@kdamas93 @ashleyvernon Here is the summary of our meeting:

1. create JSON schema for both `application_blocklist` and `application_entitlement` here,
2. run `scripts/generate-kzd-builders.escript` to generate the `kzd` accessor for `application_blocklist` and `application_entitlement` and `application`, copy them to https://github.com/2600hz/kazoo-core/pull/524
3. Do not use a static id for block list and entitlement. Let doc id be random as usual (you don't need to do anything). instead save app block or app entitlement PER app as a separate doc. store the app id in the doc
4. create a view to return application block list and entitlement per app type and app id

Second part is to document everything, (you can put the these internal docs inside `kazoo_document`, it doesn't matter where you as long as you properly document this stuff):
1. Properly document how the new application is working, what is the difference between application api and apps_store api. You mention something you copied the existing cb_apps_store code and adjusted to what you want it to do, DOCUMENT your change then.
2. same thing for init_apps and init_applications, document how they are different from each other.

If you document this properly we can review the change and see if the code is up to spec and make decision better.

priv/couchdb/schemas/application_blocklist.json
```diff
@@ -0,0 +1,27 @@
...
+    "description": "List of blocked apps for the account which queries for endpoint cb_applications",
+    "properties": {
+        "account_db": {
+            "description": "Account which the application was originally created",
+            "type": "string"
+        },
+        "account_id": {
+            "description": "Account of the application that has been blacklisted",
+            "type": "string"
+        },
```
icehess: these are private fields, and are automatically is added if you use crossbar_docs to save docs.
Do not add these fields to schema

priv/couchdb/schemas/application_blocklist.json
```diff
@@ -0,0 +1,27 @@
...
+        "application_type": {
```
icehess: rename to `type` is simpler and no need to say application again

priv/couchdb/schemas/application_blocklist.json
```diff
@@ -0,0 +1,27 @@
...
+            "description": "Application type either monster-ui or commland or some other type of wrap",
+            "type": "string"
+        },
+        "name_application": {
```
icehess: rename to `name` is simpler and no need to say application again

icehess (changes requested):

priv/couchdb/schemas/application_entitlement.json
```diff
@@ -0,0 +1,37 @@
...
+    "description": "List of entitlements related to query accounts for endpoint cb_applications",
+    "properties": {
+        "account_db": {
+            "description": "Account which the application was originally created",
+            "type": "string"
+        },
+        "account_id": {
+            "description": "The account which the entitlement was created",
+            "type": "string"
+        },
```
icehess: same thing, don't add private keys to schema

priv/couchdb/schemas/application_entitlement.json
```diff
@@ -0,0 +1,37 @@
...
+        "application_type": {
```
icehess: drop application, type is enough

priv/couchdb/schemas/application_entitlement.json
```diff
@@ -0,0 +1,37 @@
...
+        "entitlement_type": {
+            "description": "Type of entitlement granted or users with the entitlement",
+            "enum": [
+                "specific",
+                "all",
+                "admins"
+            ],
+            "type": "string"
+        },
+        "name_application": {
```
icehess: drop application, name is enough

priv/couchdb/schemas/application_entitlement.json
```diff
@@ -0,0 +1,37 @@
...
+        },
+        "entitlement_type": {
+            "description": "Type of entitlement granted or users with the entitlement",
+            "enum": [
+                "specific",
+                "all",
+                "admins"
+            ],
+            "type": "string"
+        },
```
icehess: So we want to make the entitlement to only allow setting to all users, or admins only or a list of individual users.
When the entitlement type is specific, we need another field to track the list of specific user ids.

To make this work properly we also want to make sure this entitlement type is mutually exclusive, meaning you are either to allow all/admins or allow a specific list of user ids which in turn needs and extra property call `users` to hold the list of user ids.

For achieving this mutually exclusive JSON schema provides `oneOf`, this forces only usage of only one listed schema:

```
        "entitlement": {
            "oneOf": [
                {
                    "properties": {
                        "type": {
                            "description": "Type of entitlement granted or users with the entitlement",
                            "enum": [
                                "all",
                                "admins"
                            ],
                            "type": "string"
                        }
                    },
                    "required": [
                        "type"
                    ]
                },
                {
                    "properties": {
                        "type": {
                            "description": "Type of entitlement granted or users with the entitlement",
                            "enum": [
                                "specific"
                            ],
                            "type": "string"
                        },
                        "users": {
                            "default": [],
                            "description": "Type of entitlement granted or users with the entitlement",
                            "items": {
                                "type": "string"
                            },
                            "type": "array"
                        }
                    },
                    "required": [
                        "type"
                    ]
                }
            ]
        },
```

icehess (changes requested):

priv/couchdb/schemas/application.json
```diff
@@ -0,0 +1,111 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "_id": "application",
+    "description": "Appstore applications",
+    "properties": {
+        "allowed_users": {
```
k-anderson: This is property of the entitlement not the application meta-data so should be removed.

priv/couchdb/schemas/application.json
```diff
@@ -0,0 +1,111 @@
...
+        "tags": {
+            "items": {
+                "type": "string"
+            },
+            "type": "array"
+        },
+        "urls": {
+            "type": "object"
+        },
+        "users": {
```
k-anderson: This is property of the entitlement not the application meta-data so should be removed.

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1200 @@
...
+%%%=============================================================================
+
+%%------------------------------------------------------------------------------
+%% @doc Initializes the bindings this module will respond to.
+%% @end
+%%------------------------------------------------------------------------------
+-spec init() -> 'ok'.
+init() ->
+    _ = crossbar_bindings:bind(<<"*.authenticate">>, ?MODULE, 'authenticate'),
+    _ = crossbar_bindings:bind(<<"*.authorize">>, ?MODULE, 'authorize'),
```
icehess: Remove these and their functions, these are all optional. and here you are not doing anything you can remove them then.

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1200 @@
...
+    _ = crossbar_bindings:bind(<<"*.allowed_methods.applications">>, ?MODULE, 'allowed_methods'),
+    _ = crossbar_bindings:bind(<<"*.resource_exists.applications">>, ?MODULE, 'resource_exists'),
+    _ = crossbar_bindings:bind(<<"*.content_types_provided.applications">>, ?MODULE, 'content_types_provided'),
```
icehess: Bind for content_type_accepted too and implement it. you can copy paste it from cb_apps_store. especially the part about how it handles the icons and screenshots.

Also lok for how cb_apps_Store is implementing the icon/screenshots uploads on post/put.

icehess:
Also as suggesting, if you could implement test for cb_apps_store and your new cb_application module that would be nice way to make sure we all previous features/functionalities are still present in the new API

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
+%%%-----------------------------------------------------------------------------
+%%% @copyright (C) 2011-2023, 2600Hz
+%%% @doc Endpoint for Applications
+%%% @author Kevin Damas
+%%% @end
+%%%-----------------------------------------------------------------------------
+-module(cb_applications).
+
+-export([init/0
+        , authenticate/1, authenticate/2, authenticate/3, authenticate/4, authenticate/5
```
icehess: a minor cosmetic issue, remove the space after first comma.

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+authenticate_nouns(?HTTP_PUT, [{<<"applications">>,[_AppType, _AppId, <<"entitlement">>]}]) ->'true';
+authenticate_nouns(?HTTP_POST, [{<<"applications">>,[_AppType, _AppId, <<"entitlement">>]}]) ->'true';
+authenticate_nouns(?HTTP_PATCH, [{<<"applications">>,[_AppType, _AppId, <<"entitlement">>]}]) ->'true';
+authenticate_nouns(?HTTP_DELETE, [{<<"applications">>,[_AppType, _AppId, <<"entitlement">>]}]) ->'true';
+authenticate_nouns(?HTTP_GET, [{<<"applications">>,[_AppType, _AppId, ?ICON]}]) ->'true';
+authenticate_nouns(?HTTP_POST, [{<<"applications">>,[_AppType, _AppId, ?ICON]}]) ->'true';
+authenticate_nouns(?HTTP_DELETE, [{<<"applications">>,[_AppType, _AppId, ?ICON]}]) ->'true';
+authenticate_nouns(?HTTP_GET, [{<<"applications">>,[_AppType, _AppId, ?SCREENSHOT, _Number]}]) ->'true';
+authenticate_nouns(?HTTP_POST, [{<<"applications">>,[_AppType, _AppId, ?SCREENSHOT, _Number]}]) ->'true';
+authenticate_nouns(?HTTP_DELETE, [{<<"applications">>,[_AppType, _AppId, ?SCREENSHOT, _Number]}]) ->'true';
```
icehess: Is this a dev time test thing? why authenticate the request for every API that this module provide?

Maximum thing we want to allow to open public access are these:

```erlang
authenticate_nouns(?HTTP_GET, [{<<"applications">>,[_AppType, _AppId, ?ICON]}]) ->'true';
authenticate_nouns(?HTTP_GET, [{<<"applications">>,[_AppType, _AppId, ?SCREENSHOT, _Number]}]) ->'true';
```

I can debate that we don't even need these. but for now let's do what exactly apps_store is doing.

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+authenticate_nouns(_Verb, _Nouns) ->'false'.
+
+%%------------------------------------------------------------------------------
+%% @doc Authorizes the incoming request, returning true if the requestor is
+%% allowed to access the resource, or false if not.
+%% @end
+%%------------------------------------------------------------------------------
+-spec authorize(cb_context:context()) -> boolean().
+authorize(Context) ->
+    case cb_context:is_account_admin(Context) of
```
icehess: If you can refactor this such as extract common logic `when`: reseller, master, admin to go over a common function and do pattern match on nouns and method and calls those `when` functions this makes this authorize functions more readable.

Currently it is really hard to follow :)

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+                               ]
+                              );
+        {'false', _UserId} ->
+            %% List all merged application documents that the user is assigned to.
+            %% This is used on login to load the user's commland/monsterui applications.
+            lager:warning("List all merged application documents that the user is assigned to"),
+            MergedContext = validate_merged_applications(Context, ?COMMLAND_UI, cb_context:req_verb(Context)),
+            AppsFilteredByUser = filter_entitlement_by_user(cb_context:resp_data(MergedContext), [], UserId, 'false'),
+            cb_context:set_resp_data(MergedContext, AppsFilteredByUser)
+    end;
```
icehess: This logic should be move to `cb_apps_util:allowed_applications(AppType)`, `cb_apps_util:allowed_applications(AppType, AccountId)`, and `cb_apps_util:allowed_applications(AppType, AccountId, UserId)`.

Same for next clause and anywhere in the code. The propose of this module should be only providing the API stuff (authorize, validate, basic CRUD - create, read, update, delete stuff,...) and not the business logic (how to list apps or if the account/user is allowed to access which app,...).

I suggest take a look at how `cb_apps_util:allowed_apps` is implemented and follow the same pattern. You can enhance the code but I like to see it fetches all apps from master, reseller and fold over some routines to check the app permissions, block and etc...

I can debate that we can even move this business logic back to the core. Not sure we need logic anywhere else, but just in case. Any thoughts?

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+validate(Context, _AppType) ->
+    cb_context:setters(Context,
+                       [{fun cb_context:set_resp_error_msg/2, <<"The application type doesn't not exists">>},
+                        {fun cb_context:set_resp_error_code/2, 403},
+                        {fun cb_context:set_resp_status/2, 'error'}
+                       ]
+                      ).
+
+-spec filter_entitlement_by_user (list(),list(), kz_term:ne_binary(), boolean()) -> list().
+filter_entitlement_by_user ([], Results, _UserId, _IsAdmin) ->
```
icehess: move this `cb_apps_util` we don't want this logic to be here.

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+%%------------------------------------------------------------------------------
+%% @doc Authorizes the incoming request, returning true if the requestor is
+%% allowed to access the resource, or false if not.
+%% @end
+%%------------------------------------------------------------------------------
+-spec authorize(cb_context:context()) -> boolean().
+authorize(Context) ->
+    case cb_context:is_account_admin(Context) of
+        'true' ->
+            AccountId = cb_context:account_id(Context),
```
icehess: Did you meant `cb_context:auth_account_id/1`?

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+        'success' ->
+            case cb_context:doc(Context1) of
+                [] ->
+                    Values = [{<<"name">>, ApplicationName}
+                             ,{<<"type">>, ApplicationType}
+                             ,{<<"entitlement">>, kzd_application_entitlement:entitlement(ReqData)}
+                             ,{<<"pvt_type">>, <<"application_entitlement">>}
+                             ],
+
+                    Doc = kz_json:set_values(Values, kzd_application_blocklist:new()),
```
icehess: Use module kzd_application_entitlement functions and kz_doc:setter/2 function to manipulate the doc.

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+validate(Context, ApplicationType, AppId, ?ENTITLEMENT) ->
+    validate_application_entitlement(Context, ApplicationType, AppId, cb_context:req_verb(Context)).
+
+
+-spec validate(cb_context:context(), path_token(), path_token(), path_token(), path_token()) -> cb_context:context().
+validate(Context, _AppType, AppId, ?SCREENSHOT, ScreenshotId) ->
+    validate_screenshot(Context, AppId, ScreenshotId, cb_context:req_verb(Context)).
+
+-spec validate_merged_applications(cb_context:context(), kz_term:ne_binary(), http_method()) -> cb_context:context().
+validate_merged_applications(Context, ApplicationType, ?HTTP_GET) ->
```
icehess: move this `cb_apps_util` we don't want this logic to be here.

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+validate_merged_applications(Context, ApplicationType, ?HTTP_PUT) ->
+    ReqData = cb_context:req_data(Context),
+    ApplicationName = kzd_application_entitlement:name(ReqData),
+    Context1 = crossbar_view:load(Context, ?CB_LIST, [{'key',[ApplicationType, ApplicationName]}]),
+
+    case cb_context:resp_status(Context1) of
+        'success' ->
+            case cb_context:doc(Context1) of
+                [] ->
+                    Values = [{<<"type">>, ApplicationType}
```
icehess: In general, manipulating the doc fields directly in highly discouraged. Please use standard `kz_doc` module to access common pvt fields, and use you document `kzd` module to get and set fields. That is why we created `kzd` modules in the first place to avoid these direct JObj manipulations.

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+                                        , {fun cb_context:set_doc/2, kz_json:new()}
+                                        ]
+                                      )
+            end;
+        _Status ->
+            Context1
+    end.
+
+-spec merge_application_entitlements(cb_context:context(), list(), list()) -> list().
+merge_application_entitlements(_Context, [], Results) -> Results;
```
icehess: move this `cb_apps_util` we don't want this logic to be here.

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+    case cb_context:doc(EntitlementContext) of
+        [] ->
+            Application = kz_json:set_value([<<"value">>,<<"entitlement">>], kz_json:new(), MergedApplication);
+        Entitlement ->
+            Application = kz_json:set_value([<<"value">>,<<"entitlement">>], Entitlement, MergedApplication)
+    end,
+    merge_application_entitlements(Context, MergedApplications, [Application|Results]).
+
+-spec services_apps(list(), list(), list()) -> list().
+services_apps([], _Services, Results) -> Results;
```
icehess: move this `cb_apps_util` we don't want this logic to be here.

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+    [_AppType, AppName]= kz_json:get_value(<<"key">>, MergedApplication),
+    Application = case kz_json:get_ne_binary_value(AppName, Services) of
+                      <<"block">> -> kz_json:set_value([<<"value">>,<<"published">>], 'false', MergedApplication);
+                      <<"allow">> -> kz_json:set_value([<<"value">>,<<"published">>], 'true', MergedApplication);
+                      _Else -> MergedApplication
+                  end,
+    services_apps(MergedApplications, Services, [Application|Results]).
+
+-spec blocklist_apps(list(), list(), list()) -> list().
+blocklist_apps([], _Blocklist, Results) -> Results;
```
icehess: move this `cb_apps_util` we don't want this logic to be here.

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+    crossbar_view:load(Context, ?CB_ENTITLEMENTS, Options);
+validate_application_entitlements(Context, ApplicationType, ?HTTP_PUT) ->
+    ReqData = cb_context:req_data(Context),
+    ApplicationName = kzd_application_entitlement:name(ReqData),
+    Context1 = crossbar_view:load(Context, ?CB_ENTITLEMENTS, [{'key',[ApplicationType, ApplicationName]}]),
+    case cb_context:resp_status(Context1) of
+        'success' ->
+            case cb_context:doc(Context1) of
+                [] ->
+                    Values = [{<<"name">>, ApplicationName}
```
icehess: use kzd modules

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+                              , [{fun cb_context:set_resp_status/2, 'error'}
+                                , {fun cb_context:set_resp_data/2, 'undefined'}
+                                , {fun cb_context:set_resp_error_code/2, 404}
+                                , {fun cb_context:set_resp_error_msg/2, <<"Entitlement not found">>}
+                                , {fun cb_context:set_doc/2, kz_json:new()}
+                                ]
+                              );
+        [AppEntitlementDoc] ->
+
+            Values = [{<<"name">>, ApplicationName}
```
icehess: Not gonna repeat this for the rest of code, don't manipulate JObj directly use `kzd` modules.

this is not a good place to say this, as this code can be completely done using `crossbar_doc:load_merge` and we don't ned to do anything here, follow the [`applications/skels/src/intergations/cb_crud.erl`](https://github.com/2600hz/kazoo-skel/blob/master/src/integration/cb_crud.erl) to see how this usually should be implemented.

But in general we don't manipulate the doc fields directly and we use kzd and setters. like this:
```erlang
Setters = [{fun kzd_application_entitlement:set_name/2, AppName}
          ,{fun kzd_application_entitlement:set_type/2, AppType}
          ,{fun kz_doc:set_type/2, kzd_application_entitlement:type()}
          ],
kz_doc:setters(Doc, Setters).
```

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+    case cb_context:doc(Context1) of
+        'undefined' ->
+            cb_context:setters(Context1
+                              , [{fun cb_context:set_resp_status/2, 'error'}
+                                , {fun cb_context:set_resp_data/2, 'undefined'}
+                                , {fun cb_context:set_resp_error_code/2, 404}
+                                , {fun cb_context:set_resp_error_msg/2, <<"Entitlement not found">>}
+                                , {fun cb_context:set_doc/2, kz_json:new()}
+                                ]
+                              );
```
icehess: You can return Context, it already have the proper error set.

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+        [AppEntitlementDoc] ->
+
+            Values = [{<<"name">>, ApplicationName}
+                     ,{<<"type">>, ApplicationType}
+                     ,{<<"entitlement">>, kzd_application_entitlement:entitlement(ReqData)}
+                     ],
+            Doc = kz_json:set_values(Values, kz_json:get_value(<<"doc">>, AppEntitlementDoc)),
+            cb_context:set_doc(Context1, Doc)
+    end;
+validate_application_entitlement(Context, ApplicationType, AppId, ?HTTP_PATCH) ->
```
icehess: Usually in Crossbar we impelement a common function to read the a signle doc. like here `load_application_entitlement`

Then in validate base on HTTP method/verb, we either load_application_entitlement call for GET/DELETE.

For PUT/POST/PATCH we validate the request first, meaning we validate the payload using schema, if PATCH we crossbar_doc:patch_and_validate, and then for POST after schema validation we do load_merge. All POST/PATCH/PUT wil call validate_request with the Id of doc, in case of PUT it is undefined which set the pvt using kzd_module or something or kz_doc directly. You can look at all other cb module to see how this usually is done. See and example here: https://github.com/2600hz/kazoo-skel/blob/master/src/integration/cb_crud.erl

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+                                , {fun cb_context:set_resp_error_code/2, 404}
+                                , {fun cb_context:set_resp_error_msg/2, <<"Application not found">>}
+                                , {fun cb_context:set_doc/2, kz_json:new()}
+                                ]
+                              );
+        _AppDoc ->
+            crossbar_doc:load_merge(AppId, cb_context:validate_request_data(<<"application">>, Context))
+    end;
+validate_applications(Context, ApplicationType, AppId, ?HTTP_PATCH) ->
+    ApplicationContext = crossbar_doc:load(AppId, Context),
```
icehess: See above comments, we do crossbar_doc:patch_and_validate which loads the curernt doc and merge with the req data and performs schema validation.

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+validate_screenshot(Context, AppId, ScreenshotId, ?HTTP_DELETE) ->
+    get_attachment(app_attachment_binary_meta(Context, AppId, {<<"application">>, {?SCREENSHOT, ScreenshotId}})).
+
+-spec validate_icon(cb_context:context(), kz_term:ne_binary(), http_method()) -> cb_context:context().
+validate_icon(Context, AppId, ?HTTP_GET) ->
+    ApplicationContext = crossbar_doc:load(AppId, Context),
+
+    case cb_context:doc(ApplicationContext) of
+        'undefined' ->
+            cb_context:setters(ApplicationContext
```
icehess: Like comment above please don't reset error, it ia relady set bu load function.

icehess (changes requested):

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1596 @@
...
+          cb_context:context().
+content_types_accepted(Context, ?ICON, ?HTTP_POST) ->
+    CTA = [{'from_binary', ?WHITELABEL_MIME_TYPES}],
+    cb_context:set_content_types_accepted(Context, CTA);
+content_types_accepted(Context, _AttachType, _Verb) ->
+    Context.
+
+-spec content_types_accepted(cb_context:context(), path_token(), path_token(), path_token()) -> cb_context:context().
+content_types_accepted(Context, _AppType, _AppId, ?ICON) ->
+    content_types_accepted(Context, ?ICON, cb_context:req_verb(Context)).
```
icehess: Also implement the `content_types_accepted` for icon and `sceenshots, number` and only allow image file formats like `?IMAGE_CONTENT_TYPES`

https://github.com/2600hz/kazoo-crossbar/blob/master/src/modules/cb_apps_store.erl#L169-L184

Don't forget to bind to it in init function.

icehess (changes requested):

src/cb_apps_util.erl
```diff
@@ -493,7 +706,32 @@ find_attachment(JObj, AttachmentName, [Path | Paths]) ->
...
+    {'error', 'not_found'};
+find_application_attachment(JObj, AttachmentName) ->
+    Paths = [[<<"doc">>]],
+    find_application_attachment(JObj, AttachmentName, Paths).
 
+-spec find_application_attachment(kz_json:object(), kz_term:api_ne_binary(), kz_json:paths()) ->
+          kz_either:either('not_found'
+                          ,{AttachmentName::kz_term:ne_binary(), AttachmentsObject::kz_json:object()}
+                          ).
+find_application_attachment(_, _, []) ->
```
jamesaimonetti: This function does not appear to ever have a chance to be called with `(JObj, AttachmentName, [])` as args so this clause seems unneeded.

jamesaimonetti: A question though: this function seems identical to `find_attachment`; what's the difference here?

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1035 @@
...
+
+    ContextNew = load_application(Context, AppId),
+
+    cb_context:setters(ContextNew
+                      , [{fun cb_context:set_resp_status/2, 'success'}
+                        ,{fun cb_context:set_resp_data/2, cb_context:doc(ContextNew)}]).
+
+-spec delete(cb_context:context(), path_token(), path_token(), path_token(), path_token()) -> cb_context:context().
+delete(Context, _AppType, AppId, ?SCREENSHOT,ScreenshotId) ->
+    NewObj = cb_context:doc(crossbar_doc:load(AppId,Context), ?TYPE_CHECK_OPTION(kzd_application:type())),
```
jamesaimonetti: cb_context:doc/2 does not exist. run `make xref` to show you all the places where the code makes bad function calls.

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1035 @@
...
+                                , {fun cb_context:set_resp_status/2, 'success'}
+                                ]
+                              );
+        {'error', _E} ->
+            lager:error("failed to lookup apps in ~s", [AccountId]),
+            []
+    end.
+-spec getting_account_id(cb_context:context()) -> kz_term:ne_binary().
+getting_account_id(Context) ->
+    case cb_context:account_id(Context) of
```
jamesaimonetti: cb_context:account_id does not return a tuple so why pattern match here for it?

jamesaimonetti: This whole function seems unnecessary

src/modules/cb_applications.erl
```diff
@@ -0,0 +1,1035 @@
...
+%%------------------------------------------------------------------------------
+-spec app_attachment_id(attach_type(), kz_json:object(), kz_term:ne_binary()) ->
+          kz_either:either('not_found'
+                          , {AttachmentName :: kz_term:ne_binary(), AttachmentObject :: kz_json:object()}
+                          ).
+app_attachment_id({<<"application">>, ?ICON}, Doc, _Language) ->
+    cb_apps_util:find_application_attachment(Doc, kz_json:get_ne_value([<<"doc">>, <<"icon">>], Doc));
+app_attachment_id({<<"application">>, {?SCREENSHOT, Number}}, Doc, _Language) ->
+    cb_apps_util:find_application_attachment(Doc, Number);
+app_attachment_id({<<"application">>, ?SCREENSHOT}, Doc, _Language) ->
```
jamesaimonetti: This clause does not look reachable. I don't see anywhere it is called with this first arg's tuple().

jamesaimonetti (changes requested):
Things to review and fix, see core PR review as well

icehess:
Required PRs:

- https://github.com/2600hz/kazoo-crossbar/pull/374
- https://github.com/2600hz/kazoo-core/pull/524
- https://github.com/2600hz/kazoo-properly/pull/156
- https://github.com/2600hz/kazoo-ast/pull/42

PR #377, opened 2022-11-10
[master] KZOO-316: return the correct consumed tokens buckets

icehess:
Crossbar is returning consumed tokens bucket, but it gets the consumed amount by calculating cost for default bucket which is not reflect the total amount or even considering all consumed bucket names.

We usually consumed tokens during authentication and authorization, which are implemented using `kazoo_bindings` map.

This is updating the code in `api_util` to get the all consumed tokens amount and keep from all succeeded and failed and set the result in the context so at the end we can have total amount correctly.

Also `cb_context:context()` is updated to have fields to track consumed token buckets and all used bucket names.

All modules are changed to use use `cb_modules_util` when consuming tokens, this function is properly updates the context with cost and bucket name.


icehess:
Related PRs:
* https://github.com/2600hz/kazoo-desktop/pull/19
* https://github.com/2600hz/kazoo-core/pull/526
* tests: https://github.com/2600hz/kazoo-properly/pull/105

src/api_util.erl
```diff
@@ -968,43 +975,96 @@ is_authentic_noun(Req, Context, []) ->
...
+            {'true', Req, merge_buckets(Context2, collect_remaining_buckets(T ++ Failed))};
+        [{'stop', Context2}|T] ->
             lager:debug("authn stopped"),
-            ?MODULE:stop(Req, Context2)
+            ?MODULE:stop(Req, merge_buckets(Context2, collect_remaining_buckets(T ++ Failed)))
     end.
 
--spec prefer_new_context(kz_term:proplist(), cowboy_req:req(), cb_context:context()) ->
-          {'true', cowboy_req:req(), cb_context:context()} |
+-spec prefer_new_context(kz_term:proplist(), cowboy_req:req(), cb_context:context(), 'undefined' | boolean(), kazoo_bindings:map_results()) ->
```
jamesaimonetti: last arg is not map_results() but bucket_collection() right?

icehess:
backport PRs:

5.1: https://github.com/2600hz/kazoo-crossbar/pull/394
5.0: https://github.com/2600hz/kazoo-crossbar/pull/395

PR #381, opened 2022-11-15
KCAL-45 Add aliases to the voicemail box 

nkarunarathne:
Previously the voicemail box had only a voicemail box number for box identification. It was requested add new aliases so that the voicemail box can be looked up using either the voicemail number or aliases.

The new 'aliases' parameter is a list of strings that can be used to identify a voicemail box uniquely. Each voicemail box number and/or aliases are always unique in the account.

Also, The voicemail aliases can be used as transfer targets. The voicemail aliases can be used when calling to check voicemail where the voicemail box number is asked for.

The Code Level Change

Edited the existing listing_by_mailbox view so that it would check both voicemail box IDs and aliases when doing the lookups.
Check the uniqueness of both voicemail box IDs and aliases when receiving API requests.
The voicemail box lookup was using an integer box number, changed it to binary. 

Changes are in
crossbar, callflow, core


src/modules/cb_vmboxes.erl
```diff
@@ -1327,3 +1327,7 @@ update_mwi(Context, BoxId, 'success') ->
     Context;
 update_mwi(Context, _BoxId, _Status) ->
     Context.
+
+get_aliases_and_mailbox(Context)->
+    [kz_json:get_integer_value(<<"mailbox">>, cb_context:req_data(Context))
+    | kz_json:get_value(<<"aliases">>, cb_context:req_data(Context), [])].
```
icehess:
```suggestion
    [kzd_vmboxes:mailbox(cb_context:req_data(Context))
     | kzd_vmboxes:aliases(cb_context:req_data(Context), [])
    ].
```

src/modules/cb_vmboxes.erl
```diff
@@ -1273,24 +1273,25 @@ generate_media_name(CallerId, GregorianSeconds, Ext, Timezone) ->
...
-        Mailbox ->
-            check_uniqueness(VMBoxId, Context, Mailbox)
+    try get_aliases_and_mailbox(Context) of
+        BoxIdentities ->
+            check_uniqueness(VMBoxId, Context, BoxIdentities)
     catch
         _:_ ->
             lager:debug("can't convert mailbox number to integer"),
             'false'
     end.
```
icehess: We don't need try/cactch anymore, so you can get ride of it.
Also handle empty case!

```
    case get_aliases_and_mailbox(Context) of
        [] ->
            lager:debug("no mailbox number or aliases"),
            'false';
        BoxIdentities ->
            check_uniqueness(VMBoxId, Context, BoxIdentities)
    end.
```

icehess (changes requested):
You also need to modify these crossbar views, search for vmbox cases and add aliases to mailbox if it needs:

* `core/kazoo_couch/priv/couchdb/views/account-crossbar_listings.json` in kazoo-core
* `applications/locations/priv/couchdb/views/account-locations.json` in kazoo-locations repo

src/modules/cb_vmboxes.erl
```diff
@@ -1327,3 +1327,8 @@ update_mwi(Context, BoxId, 'success') ->
     Context;
 update_mwi(Context, _BoxId, _Status) ->
     Context.
+
+get_aliases_and_mailbox(Context)->
+    [kzd_vmboxes:mailbox(cb_context:req_data(Context))
+    | kzd_vmboxes:aliases(cb_context:req_data(Context), [])
+    ].
```
icehess: fmt is erroring because we prefer to have new line  character at the end of file, just hit enter after `.` here will fix it.

src/modules/cb_vmboxes.erl
```diff
@@ -1327,3 +1327,8 @@ update_mwi(Context, BoxId, 'success') ->
     Context;
 update_mwi(Context, _BoxId, _Status) ->
     Context.
+
+get_aliases_and_mailbox(Context)->
+    [kzd_vmboxes:mailbox(cb_context:req_data(Context))
+    | kzd_vmboxes:aliases(cb_context:req_data(Context), [])
```
icehess: I could be wrong but i think our formatter likes a space before `|`:

```suggestion
    [kzd_vmboxes:mailbox(cb_context:req_data(Context))
     | kzd_vmboxes:aliases(cb_context:req_data(Context), [])
```

src/modules/cb_vmboxes.erl
```diff
@@ -1299,9 +1299,9 @@ check_uniqueness(VMBoxId, Context, Mailbox) ->
         {'ok', []} -> 'true';
         {'ok', [VMBox]} ->
             VMBoxId =:= kz_doc:id(VMBox);
-        {'ok', _} ->
-            lager:warning("found multiple mailboxes for '~p'", [Mailbox]),
-            'false';
+        {'ok', Boxes} ->
+            (lists:all(fun(VMBox) -> VMBoxId =:= kz_doc:id(VMBox) end, Boxes)) andalso
+            (length(BoxIdentities) =:= length(props:unique(BoxIdentities)));
```
icehess: We usually drop `and/or` or `andalso/orelse` to a new line followed by right hand side, also no need for parentheses:

```suggestion
            lists:all(fun(VMBox) -> VMBoxId =:= kz_doc:id(VMBox) end, Boxes)
                andalso length(BoxIdentities) =:= length(props:unique(BoxIdentities));
```

icehess (changes requested):

src/modules/cb_vmboxes.erl
nkarunarathne: There is no space there, it is usually like this 

https://github.com/2600hz/kazoo-ast/blob/2af19c3f246efe3ae78069821149c3c8d793384c/src/cb_api_endpoints.erl#L909-L912

src/modules/cb_vmboxes.erl
icehess: Let's do this instead:
```suggestion
    lists:usort(
      [BoxIdentity
       || BoxIdentity <- [kzd_vmboxes:mailbox(cb_context:req_data(Context))
                         | kzd_vmboxes:aliases(cb_context:req_data(Context), [])
                         ],
          kz_term:is_not_empty(BoxIdentity)
      ]
    ].

```

src/modules/cb_vmboxes.erl
```diff
@@ -1299,9 +1299,9 @@ check_uniqueness(VMBoxId, Context, Mailbox) ->
         {'ok', []} -> 'true';
         {'ok', [VMBox]} ->
             VMBoxId =:= kz_doc:id(VMBox);
-        {'ok', _} ->
-            lager:warning("found multiple mailboxes for '~p'", [Mailbox]),
-            'false';
+        {'ok', Boxes} ->
+            lists:all(fun(VMBox) -> VMBoxId =:= kz_doc:id(VMBox) end, Boxes)
+                andalso length(BoxIdentities) =:= length(props:unique(BoxIdentities));
```
icehess, line 1304: hmm, I'm not following what this line is checking exactly?

src/modules/cb_vmboxes.erl
```diff
@@ -1273,24 +1273,24 @@ generate_media_name(CallerId, GregorianSeconds, Ext, Timezone) ->
...
 check_uniqueness(VMBoxId, Context) ->
-    try kz_json:get_integer_value(<<"mailbox">>, cb_context:req_data(Context)) of
-        Mailbox ->
-            check_uniqueness(VMBoxId, Context, Mailbox)
-    catch
-        _:_ ->
-            lager:debug("can't convert mailbox number to integer"),
-            'false'
+    case get_aliases_and_mailbox(Context) of
+        ['undefined'] ->
```
icehess: With the change suggested for `get_aliases_and_mailbox` you can safely check for empty list here.

src/modules/cb_vmboxes.erl
```diff
@@ -1273,24 +1273,24 @@ generate_media_name(CallerId, GregorianSeconds, Ext, Timezone) ->
...
+            lager:debug("no mailbox number or aliases"),
+            'false';
+        BoxIdentities ->
+            check_uniqueness(VMBoxId, Context, BoxIdentities)
     end.
 
--spec check_uniqueness(kz_term:api_binary(), cb_context:context(), pos_integer()) -> boolean().
-check_uniqueness(VMBoxId, Context, Mailbox) ->
-    ViewOptions = [{'key', Mailbox}],
+-spec check_uniqueness(kz_term:api_binary(), cb_context:context(), kz_term:api_list()) -> boolean().
```
icehess:
```suggestion
-spec check_uniqueness(kz_term:api_binary(), cb_context:context(), kz_term:ne_binaries()) -> boolean().
```

icehess (changes requested):
lgtm, just some small changes.

src/modules/cb_vmboxes.erl
nkarunarathne: Why do we need to sort this, order dose not matter here I think.

src/modules/cb_vmboxes.erl
nkarunarathne: This check will prevent adding the same VM box number or alias to other boxes as well as, the same VM box number to the same box aliases list as well.

src/modules/cb_vmboxes.erl
icehess: `usort` is sort and **unique**, so you don't need to unique it in after you fetch the view.

nkarunarathne: got it

nkarunarathne: But if I do that It will break the logic in line 1303.

src/modules/cb_vmboxes.erl
```diff
@@ -1327,3 +1327,11 @@ update_mwi(Context, BoxId, 'success') ->
     Context;
 update_mwi(Context, _BoxId, _Status) ->
     Context.
+
+get_aliases_and_mailbox(Context)->
+    [BoxIdentity
+     || BoxIdentity <- [kzd_vmboxes:mailbox(cb_context:req_data(Context))
+                       | kzd_vmboxes:aliases(cb_context:req_data(Context), [])
+                       ],
+        kz_term:is_not_empty(BoxIdentity)
```
danielfinke, line 1336: If you move the check for uniqueness to after the schema validation, you won't need to check if the identities are empty (since the schema is checking `"minLength": 1` for each alias

src/modules/cb_vmboxes.erl
```diff
@@ -785,12 +785,12 @@ validate_unique_vmbox(VMBoxId, Context) ->
 validate_unique_vmbox(VMBoxId, Context, 'undefined') ->
     check_vmbox_schema(VMBoxId, Context);
 validate_unique_vmbox(VMBoxId, Context, _AccountDb) ->
-    case check_uniqueness(VMBoxId, Context) of
-        'true' -> check_vmbox_schema(VMBoxId, Context);
+    ValidatedContext = check_vmbox_schema(VMBoxId, Context),
+    case check_uniqueness(VMBoxId, ValidatedContext) of
```
icehess: Now since you are checking schema first, then before checking the uniqueness you need to check if `validateContext` resp_status is `success` then continue with uniqueness.

Something like this would suffice:
```
    case cb_context:resp_status(ValidatedContext) of
        'success' ->
                case check_uniqueness(VMBoxId, ValidatedContext) of
....
....
        _SchemaInvalid ->
            ValidateContext
    end.
```

Other option is to move check uniqueness to `OnSuccess` function of validate schema. but you need to maintain (or actually move) the check db name concept of `validate_unique_vmbox` to the new function.

icehess (changes requested):

src/modules/cb_vmboxes.erl
jamesaimonetti: Interestingly, the properly check I added failed when trying to create a second box with the same "mailbox" as first created, suggesting the uniqueness check, as-is, isn't working as expected. See https://app.circleci.com/pipelines/github/2600hz/kazoo-properly/571/workflows/e7e6a9da-3424-4b14-991f-25b34e45b889/jobs/4174

icehess: lol, probably it couldn't get mailbox properly because of invalid schema?

PR #383, opened 2022-11-15
[5.1] - KCAL-46: Voicemail envelope prompts customization to enable customized prompt

sanjeevie6353:
[5.1] KCAL-46 : Voicemail envelope prompts customization to enable customized prompt

Previously, the voicemail envelope only allowed enabling and disabling the envelope to play or ignore the envelope part of the voicemail message announcement. It just played the date and time of the voicemail occurrence.

This new feature will allow to select a specific envelope prompt based on the need. It will be configurable per voicemailbox.
This implementation  allow to either skip envelope,  play envelope to announce time of voicemail or play envelope to announce caller and the time of voicemail.


nkarunarathne:
Reviewed the changes

PR #384, opened 2022-11-16
KCRO-163: View fax pdf inline in web browser

Navoda626:
Previously, when trying to view a fax attachment it will open a new tab in browser but it will response as 404 NOT found message.

Now the header content-disposition of attachement response was changed to inline instead of attachment, so that it will display on a new tab in the browser.


lazedo:
@Navoda626 what if we want to download the attachment ? you're fixing one case but creating another.

Navoda626:
> @Navoda626 what if we want to download the attachment ? you're fixing one case but creating another.

@lazedo Yes there's a concern. Decided to call API with different query parameter values for viewing and downloading

PR #386, opened 2022-11-22
KZOO-323: add support for user-defined metadata on call recordings

danielfinke:
- `user_metadata` can be modified with a POST to `/recordings`
- attempt to update metadata in storage backends before saving in
  crossbar
- doc usage of metadata for S3 object tag-set when S3 attachment handler
  is configured in storage plan

kazoo5 PR: https://github.com/2600hz/kazoo5/pull/87
kazoo-core PR: https://github.com/2600hz/kazoo-core/pull/536
kazoo-core PR 2: https://github.com/2600hz/kazoo-core/pull/550
kazoo-media-mgr PR: https://github.com/2600hz/kazoo-media-mgr/pull/29


src/modules/cb_recordings.erl
```diff
@@ -112,9 +114,36 @@ validate_recording(Context, RecordingId, ?HTTP_GET) ->
...
+%% @doc Save changes to the writable properties of a recording. If there have
+%% been no changes, this is a no-op.
+%% @end
+%%------------------------------------------------------------------------------
+-spec post(cb_context:context(), path_token()) -> cb_context:context().
+post(Context, RecordingId) ->
+    Doc = cb_context:doc(Context),
+    Diff = cb_context:fetch(Context, 'diff'),
+
+    case kz_json:is_json_object(kzd_call_recordings:key_user_metadata(), Diff) of
```
icehess: How about using the accessor directly?

```suggestion
    case kz_json:is_json_object(kzd_call_recordings:user_metadata(Diff)) of
```

icehess (changes requested):

src/modules/cb_recordings.erl
```diff
@@ -286,3 +315,68 @@ acceptable_content_types(Context) ->
...
+    cb_context:set_req_data(Context, ReqData1).
+
+%%------------------------------------------------------------------------------
+%% @doc Validate that only writable properties of the recording have been
+%% changed in the request. Errors are added if read-only properties have been
+%% changed.
+%% @end
+%%------------------------------------------------------------------------------
+-spec validate_only_writable_props_changed(cb_context:context()) -> cb_context:context().
+validate_only_writable_props_changed(Context) ->
```
jamesaimonetti: I think once you go PATCH, just verify that only write-able fields are included in the request data. If any field is not write-able, schema error returned.

Then you don't have to do this funky diff dance below.

danielfinke:
@jamesaimonetti ready for re-review and also opened up this one -> https://github.com/2600hz/kazoo-core/pull/550

src/modules/cb_recordings.erl
```diff
@@ -112,9 +123,42 @@ validate_recording(Context, RecordingId, ?HTTP_GET) ->
...
 
+%%------------------------------------------------------------------------------
+%% @doc Validate the request body of a `PATCH' request.
+%% @end
+%%------------------------------------------------------------------------------
+-spec validate_patch(path_token(), cb_context:context()) -> cb_context:context().
+validate_patch(_, Context) ->
+    cb_context:validate_request_data(<<"call_recordings">>
+                                    ,strip_id_from_req_data(Context)
+                                    ,fun validate_only_writable_props_included/1
```
jamesaimonetti: I think you want to validate the req data before you call `crossbar_doc:patch_and_validate/4` above, since we would want to error the request is a non-writeable field is present in the request.

As-is, you fetch the doc, merge the request, validate the merged doc, then check the request for bad fields - a lot of unnecessary work

danielfinke: Good point, I've switched the order of those validations

PR #387, opened 2022-11-23
KCRO-167: Streamchat Kazoo integration not updating existing users

kdamas93:


src/modules/cb_getstream.erl
```diff
@@ -237,8 +237,28 @@ create_getstream_url(Resource, QS) ->
 -spec check_getstream_body(kz_term:api_list(), kz_term:api_binary(), cb_context:context(), kz_term:api_ne_binary(), kz_term:api_binary(), kz_json:api_json_term()) -> boolean().
 check_getstream_body([], UserId, Context, APIKey, Secret, UserData) ->
     create_getstream_user(UserId, Context, APIKey, Secret, UserData);
-check_getstream_body(_User, _UserId, _Context, _APIKey, _Secret, _UserData) ->
-    'true'.
+check_getstream_body([User], UserId, Context, APIKey, Secret, UserData) ->
+    RoleGestream = kz_json:get_ne_value(<<"role">>, User),
+    UsernameGetstream =kz_json:get_ne_value(<<"username">>, User),
+    NameGetstream = kz_json:get_ne_value(<<"name">>, User),
```
icehess: It is better to use `kz_json:get_ne_binary_value/2` so we know they should be string.

src/modules/cb_getstream.erl
```diff
@@ -237,8 +237,28 @@ create_getstream_url(Resource, QS) ->
...
+
+    Role = kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>], <<"priv_level">>], UserData),
```
icehess: Just checking is this correct? how can we assume user_id can be priv_level?

icehess: Also for this and the lines below it, is the order of keys correct? Isn't better to first check for `username` if it is missing then use user_id?

src/modules/cb_getstream.erl
```diff
@@ -237,8 +237,28 @@ create_getstream_url(Resource, QS) ->
...
+    UserName = kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>], <<"username">>], UserData),
+
+    FirstName = kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>],<<"first_name">>],UserData),
+    LastName = kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>],<<"last_name">>],UserData),
+    Name = <<FirstName/binary, LastName/binary>>,
```
icehess: Just checking do we expect getstream also concat first and last names without any space between them?

icehess (changes requested):
Some code change suggestions.

Also fyi alternative approach is to there separate functions one each to check if getstream role, username and name are matching with request.

src/modules/cb_getstream.erl
```diff
@@ -273,6 +294,42 @@ create_getstream_user(UserId, Context, APIKey, Secret, UserData)->
...
+
+    {_, TokenHeader} = create_token_for_getstream('header', <<"write">>, Secret),
+
+    Headers = create_headers(TokenHeader),
+
+    FirstName = kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>],<<"first_name">>],UserData),
+    LastName = kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>],<<"last_name">>],UserData),
+
+    Params = kz_json:from_list_recursive(
+               [{<<"users">>, [[{<<"id">>, UserId}
```
jamesaimonetti: It can be difficult to parse large bodies like this and make changes at the right level of the object.

I would suggest breaking the pieces out (like the `set` value should just be an object created before Params and included as `{<<"set">>, SetParams}`

src/modules/cb_getstream.erl
```diff
@@ -273,6 +294,42 @@ create_getstream_user(UserId, Context, APIKey, Secret, UserData)->
...
+                               ]
+                              ]
+                }
+               ]
+              ),
+
+    JsParams = kz_json:encode(Params),
+
+    case kz_http:patch(URL, Headers, JsParams) of
+        {'ok', 200, _RespHeaders, RespBody} ->
```
jamesaimonetti: If you aren't using a value, or only using it for logging purposes, prefix with the variable with `_`.

src/modules/cb_getstream.erl
kdamas93: changes added : 
```
 RoleGetstream = kz_json:get_ne_binary_value(<<"role">>, User),
 UsernameGetstream =kz_json:get_ne_binary_value(<<"username">>, User),
 NameGetstream = kz_json:get_ne_binary_value(<<"name">>, User),
```

src/modules/cb_getstream.erl
kdamas93: > Just checking is this correct? how can we assume user_id can be priv_level?

I do not fully understand the question, it was defined as follows at the beginning, in which case the getstream section does not exist, it will take the kazoo one, which is the priv_level that defines the role in kazoo, be it `admin` or `user`

kdamas93: > Also for this and the lines below it, is the order of keys correct? Isn't better to first check for `username` if it is missing then use user_id?

Actually the purpose is to compare if any property has changed to make the `patch`, therefore in my opinion I consider that the order does not matter

src/modules/cb_getstream.erl
kdamas93: > Just checking do we expect getstream also concat first and last names without any space between them?

Yes, since that was the way we created it, I attach a code snippet

kdamas93: ` 

    Headers = create_headers(TokenHeader),

    FirstName = kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>],<<"first_name">>],UserData),
    LastName = kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>],<<"last_name">>],UserData),

    Body = kz_json:from_list(
             [{<<"id">>, UserId},
              {<<"teams">>, [cb_context:account_id(Context)]},
              {<<"username">>, kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>], <<"username">>], UserData)},
              {<<"name">>, <<FirstName/binary, LastName/binary>>},
              {<<"role">>, kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>], <<"priv_level">>], UserData)}
             ]),
`

kdamas93: This is located in the function `create_getstream_user`

src/modules/cb_getstream.erl
kdamas93: Changes added: 
```
UserProperties = {<<"set">>, [{<<"username">>, kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>], <<"username">>], UserData)}
                                 ,{<<"name">>, <<FirstName/binary, LastName/binary>>}
                                 ,{<<"role">>, kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>], <<"priv_level">>], UserData)}
                                 ]
                     },
```

```
Params = kz_json:from_list_recursive(
               [{<<"users">>, [[{<<"id">>, UserId}
                               ,UserProperties
                               ]
                              ]
                }
               ]
              ),
```

src/modules/cb_getstream.erl
kdamas93:
```
case kz_http:patch(URL, Headers, JsParams) of
        {'ok', 200, _RespHeaders, _RespBody} ->
            lager:debug("user updated: ~p",[_RespBody]),
            'true';
        {'error', _E} ->
            lager:debug("failed: ~p", [_E]),
            'false'
    end
```

src/modules/cb_getstream.erl
icehess: @kdamas93 Try this:

```
    Usernmae = kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>], <<"username">>], UserData),
    UserRole = kz_json:get_first_defined([[<<"getstream">>, <<"user_id">>], <<"priv_level">>], UserData),
    UserProperties = kz_json:from_list(
                       [{<<"id">>, UserId}
                       ,{<<"set">>
                        ,kz_json:from_list(
                           [{<<"username">>, Username}
                           ,{<<"name">>, <<FirstName/binary, LastName/binary>>}
                           ,{<<"role">>, UserRole}
                           ]
                          )
                        }
                       ]
                      ),

    Params = kz_json:from_list([{<<"users">>, [UserProperties]}]),
```

src/modules/cb_getstream.erl
icehess: You are using get_first_defined, what if `[getstream, user-id]` is not changed but the user's priv_level changed on kazoo side? So the order is matter!

Then, first of all, you are using `[getstream, user_id]` for everything from username, to priv_level to name and etc.

I don't know what is `[getstream, user-id]` or where it came from or is this something that user can set by themselves or it is changeable by user at all.  I don 't even know even what getstream is. But for me the logic of using the same key for everything seems wrong.

What if there is indeed `[getstream, user-id]` key define in the user doc? then we are using the same key for everything, such as name, username and role.

If you want to get the role, then get it from `priv_level`. we always have priv_level, first_name and last_name and username define in user doc.

If you do fist get getstream user_id then fall back to to other keys, then ahen we figureout that user_id in the future we never use the other key. Like when you are calling update getstream it always use user_id for everything which is wrong and can replace the correct value of the role for example when the user first initially create in getstream.

Also imo we should NOT allow allow changing the getstream user_id. the proper way is to save it as a private key and leak the value as readonly to crossbar response.

I know this is not what you're trying to fix here, but the logic here seems wrong.

In other hand, if the indent here is to check if any of those fields are changed, then instead of doing this here you can simply create a list of all key paths that you want to check and fold over it and compare original and new one to see if they changed.

icehess (changes requested):

kdamas93:
You are using get_first_defined, what if [getstream, user-id] is not changed but the user's priv_level changed on kazoo side? So the order is matter!

**KD**: Order changed.
`UserRole = kz_json:get_first_defined([<<"priv_level">>, [<<"getstream">>, <<"role">>]], UserData),`

Then, first of all, you are using [getstream, user_id] for everything from username, to priv_level to name and etc.

I don't know what is [getstream, user-id] or where it came from or is this something that user can set by themselves or it is changeable by user at all. I don 't even know even what getstream is. But for me the logic of using the same key for everything seems wrong.

**KD:** Observation understood:
Change made

```
FirstName = kz_json:get_first_defined([[<<"getstream">>, <<"first_name">>],<<"first_name">>],UserData),
LastName = kz_json:get_first_defined([[<<"getstream">>, <<"last_name">>],<<"last_name">>],UserData),

Username = kz_json:get_first_defined([[<<"getstream">>, <<"username">>], <<"username">>], UserData),
```


What if there is indeed [getstream, user-id] key define in the user doc? then we are using the same key for everything, such as name, username and role.
**KD :** Changes made :
`UserId = kz_json:get_value(<<"_id">>, UserData),`


If you want to get the role, then get it from priv_level. we always have priv_level, first_name and last_name and username define in user doc.

If you do fist get getstream user_id then fall back to to other keys, then ahen we figureout that user_id in the future we never use the other key. Like when you are calling update getstream it always use user_id for everything which is wrong and can replace the correct value of the role for example when the user first initially create in getstream.

Also imo we should NOT allow allow changing the getstream user_id. the proper way is to save it as a private key and leak the value as readonly to crossbar response.

I know this is not what you're trying to fix here, but the logic here seems wrong.

In other hand, if the indent here is to check if any of those fields are changed, then instead of doing this here you can simply create a list of all key paths that you want to check and fold over it and compare original and new one to see if they changed.
**KD:**
@icehess  I hadn't realized that I was using the **[getstream, user_id]** for all the keys
The requested changes have been made, I appreciate your patience and explanation, let me know if there is anything pending

icehess:
@kdamas93 I refactored the code a bit. please check the code for any semantic error or any formatting issues and etc.

the summary of changes:

* instead of passing so many arguments to function, you can simply pass the minimum argument required, here we if pass context around, we automatically have access to everything we need. This way the code is more readable.
* make simple and beautiful :))) functions that it is clear what they are doing, and give it a good name too.
* A good practice is , try to think of the logic that you are trying to implement, and see if you can encapsulate some inner of logic into smaller function, so they don't make the bigger logic hard to follow and understand. If you follow the refactored code, you see all functions related that are dealing with making request to getstream to out of main functions that deciding what do.
* Another good practice is to try and reuse code if possible, se how in the new changes related to getting and constructing the get stream field is re-used, or how the req header or url functions are re-used.
* When dealing with API, always try to return meaningful errors
* if possible try and use `kazoo_document` modules, like here we can use `kzd_users` to access user's fields

kdamas93:
@icehess I made a few small changes that I noticed were missing while testing,
thanks in advance for the guide

src/modules/cb_getstream.erl
icehess: Do not log RespBody

src/modules/cb_getstream.erl
```diff
@@ -201,103 +199,187 @@ maybe_get_default_secret(_Context) ->
...
-create_getstream_user(UserId, Context, APIKey, Secret, UserData)->
-    QS = [{<<"api_key">>, APIKey}],
-    URL = create_getstream_url(<<"users">>, QS),
+    URL = build_url(Context, []),
+    Headers = create_headers(Context, <<"write">>),
 
-    {_, TokenHeader} = create_token_for_getstream('header', <<"write">>, Secret),
+    case kz_http:post(URL, Headers, ReqBody) of
+        {'ok', 201, _RespHeaders, RespBody} ->
+            lager:debug("getstream user created: ~p",[RespBody]),
```
icehess: Do not log RespBody

icehess (changes requested):

src/modules/cb_getstream.erl
kdamas93: changes made!

icehess (approved):

PR #391, opened 2022-11-28
KZOO-324 : Adding possibility to configure default custom application variables to appear on CDRs

sanjeevie6353:
KZOO-324 : Adding possibility to configure default custom application variables to appear on CDRs

Previously, the CDRs were printed with **custom application variables** only when the user has assigned them. 

This new feature will allow to define a set of  **custom application variables** in a system config. These defined sys.config values will always be available in the CDRs.


sanjeevie6353:
Closing the PR as the change on this PR is already merged onto master.

PR #393, opened 2022-11-29
Fix schemas API error

harenson:
Since the second parameter to `cb_schemas:auth_nouns/2` is a proplist with 2 elements ([{<<"schemas">>, _}, {<<"account">>, _}]), it is crashing every time the function is called because the function's head is expecting a 1 element proplist.


lazedo:
why are you calling with `account_id` ?

harenson:
@lazedo My bad, I did not know it could be queried without providing an account. Thank you 👍🏾

jamesaimonetti:
@harenson please don't take it upon yourself to "fix" things like this without doing the bare minimum to check if the code is as it is on purpose. A simple check of the doc at https://github.com/2600hz/kazoo-crossbar/blob/master/doc/schemas.md would have shown you the expected way to call the API and saved you time. If the docs don't clear it up for you, the next step is to ask in training sessions (or in chat) what the intent of the API is.

Let's not invent problems when we have enough to handle already! :p

harenson:
@jamesaimonetti Makes sense, thank you for the tips :)

PR #397, opened 2022-12-05
KZOO-222: only authorize known requests

jamesaimonetti:
Previously, a change had been made to bind for the 'authorize' request and, if cb_multi_factor didn't handle the authorization, the module would return 'true'.

This permitted unauthorized requests to proceed since api_util is checking for any 'true' results to the 'authorize' check.

Any module handling the 'authorize' callback must return 'false' unless the explicit case for handling a request is met.


jamesaimonetti:
Test for child access to ancestor accounts: https://github.com/2600hz/kazoo-properly/pull/112

PR #399, opened 2022-12-06
KCRO-164: Allow to send payloads to pusher enabled devices

harenson:
Depends on https://github.com/2600hz/kazoo-core/pull/578

Adds 2 new APIs to allow payloads to be sent to pusher enabled/configured devices via pusher application:

- `PUT /users/{USER_ID}/push`: Send payload to given device.
- `PUT /devices/{DEVICE_ID}/push`: Send payload to user's devices.


doc/ref/devices.md
```diff
@@ -426,6 +426,16 @@ curl -v -X GET \
     http://{SERVER}:8000/v2/accounts/{ACCOUNT_ID}/devices/status
 ```
 
+## Create
+
+> PUT /v2/accounts/{ACCOUNT_ID}/devices/{DEVICE_ID}/push
+
```
jamesaimonetti: These probably aren't needed now, right?

harenson: I guess those could be deleted, or just updated to match the new _pushes_ URI.

harenson: Updated _users_ and _devices_ push-related document to point to _pushes_ documentation file.

jamesaimonetti: The ref docs' sections on /push should be deleted now. Why are they still in the diff here?

harenson:
@lazedo All the checks but _docs_ are passing now. _docs_ check is expected to fail due to the custom Crossbar API implementation and the auto-generated documentation difference.

PR #400, opened 2022-12-09
KCRO-167: Streamchat Kazoo integration not updating existing users (#…

kdamas93:
…387)

* KCRO-167: Streamchat Kazoo integration not updating existing users

With this change, updates are made to existing users (patch), the patch is only made when one of the fields has been updated.

* removing unnecessary logs


src/modules/cb_getstream.erl
```diff
@@ -201,103 +199,185 @@ maybe_get_default_secret(_Context) ->
...
-validate_if_user_exist_on_getstream(UserId, Context, Secret, APIKey, UserData) ->
-    {_, TokenHeader} = create_token_for_getstream('header' ,<<"read">>, Secret),
-
-    Headers = create_headers(TokenHeader),
-
-    Eq = kz_json:from_list([{<<"$eq">>,UserId}]),
-    Id = kz_json:from_list([{<<"id">>, Eq}]),
-    Params = kz_json:from_list([{<<"filter_conditions">>, Id}]),
-    JsParams = kz_json:encode(Params),
+-spec create_or_update_user_on_getstream(cb_context:context()) -> kz_either:either(any(), kz_json:object()).
```
jamesaimonetti: bad spec return

jamesaimonetti (changes requested):

src/modules/cb_getstream.erl
kdamas93: Changes made ! @jamesaimonetti  , Thank you !

PR #401, opened 2022-12-13
KBLA-11: Add missed_call event handling

wdhammika:
Previously, There were no missed call notifications sent to the client's IOS/Android device/s using the pusher.
Now, A push notification will deliver to the Client's IOS/Android device when there is a missed call.

Preconditions: The device should have the push capability and it should be enabled.
An application should have registered in pusher in order to capture push requests on the device.

Missed call alerts should be enabled from the call flow.
(This PR is relevant to missing configurations for the kbla-11)

Related Changes on - Crossbar, Blackhole


nkarunarathne:
I reviewed the changes, seems ok to me.

PR #402, opened 2022-12-14
KCRO-169: Save account doc when resetting API key

jamesaimonetti:
Previously, only the account doc in the 'accounts' db would be saved with the new API key.

This change ensures the account doc in the account db is updated first, followed by the 'accounts' version.


jamesaimonetti:
See https://github.com/2600hz/kazoo-properly/pull/113 for test

PR #403, opened 2022-12-15
KCRO-149: Implementation of Websites API

Navoda626:
KCRO-149: Implementation of Websites API

Implementation of websites API to facilitate adding websites using the comm.land. So an admin and users would be able to add custom websites with a logo.

A new document with the type 'website' is created in the database for every website added. 'name' and the 'web_url' are mandatory parameters for a website document and there is 'users' parameter which holds the list of users whom the website is assigned. And there is a 'account_wide' parameter works as a flag to assign the website to every user in that account.

Create, update, delete and read APIs implemented for the websites. 

Additionally a binary logo can be updated to a website document and fetched from a website document.


doc/websites.md
```diff
@@ -0,0 +1,297 @@
...
+
+## About Websites
+
+Information about the websites in an account and how those websites are assigned to users.
+
+## Fetch Available Website Bindings
+
+List all available websites in an account
+
+> GET /v2/websites
```
jamesaimonetti: Fix

doc/websites.md
```diff
@@ -0,0 +1,297 @@
...
+
+```shell
+curl -v -X GET \
+    -H "Content-Type:application/json" \
+    -H "X-Auth-Token: {AUTH_TOKEN} \
+    http://{SERVER}:8000/vs/accounts/{ACCOUNT_ID}/websites
```
jamesaimonetti: fix version in url

doc/websites.md
```diff
@@ -0,0 +1,297 @@
...
+**Response**
+
+```json
+{
+    "data": [
+        {
+            "id": "e4a04b3e0be66cace9d773c2b7931101",
+            "name": "twitter",
+            "web_url": "http://www.twitter.com",
+            "logo": "dgsvsvasvaav",
```
jamesaimonetti: what is the value supposed to represent? this value doesn't help me know.

doc/websites.md
```diff
@@ -0,0 +1,297 @@
...
+    "metadata": {
+        "id": "7b5670335c21409ba33d78bbbc708ff1",
+        "created": 63837101837,
+        "modified": 63837580528
+    },
+    "status": "success"
+}
+```
+
+## GET a website logo
```
jamesaimonetti, line 294: why a GET to /logo when you can GET to /{WEBSITE_ID} with accept header?

priv/couchdb/schemas/websites.json
```diff
@@ -0,0 +1,39 @@
...
+        "account_wide": {
+            "default": true,
+            "description": "Whether an website is visible to all accounts in the parent account",
+            "type": "boolean"
+        },
+        "name": {
+            "description": "A name for the website to be displayed",
+            "maxLength": 128,
+            "minLength": 1,
+            "support_level": "supported",
```
jamesaimonetti: don't add support_level to properties; this is more of the support team's task to do once 2600Hz "officially" supports something

doc/websites.md
```diff
@@ -0,0 +1,297 @@
+# Websites
+
+## About Websites
+
+Information about the websites in an account and how those websites are assigned to users.
```
jamesaimonetti: Can you write about why websites are needed / used in KAZOO? Why would someone want or need to use this API?

src/modules/cb_websites.erl
```diff
@@ -0,0 +1,445 @@
...
+    'ok'.
+
+%%------------------------------------------------------------------------------
+%% @doc Add content types accepted and provided by this module
+%% @end
+%%------------------------------------------------------------------------------
+
+-spec acceptable_content_types() -> [cowboy_content_type()].
+acceptable_content_types() ->
+    ?IMAGE_CONTENT_TYPES.
```
jamesaimonetti: Should this be `?WEBSITE_LOGO_MIME_TYPES`? or is this only for content types provided by this module?

src/modules/cb_websites.erl
```diff
@@ -0,0 +1,445 @@
...
+-spec load_summary(cb_context:context()) -> cb_context:context().
+load_summary(Context) ->
+    case cb_context:req_nouns(Context) of
+        [{<<"websites">>, []}, {<<"users">>, [UserId]} |_] ->
+            user_summary(Context, UserId);
+        [{<<"websites">>, []}, {<<"accounts">>, [_AccountId]} |_] ->
+            lager:debug("getting account summary"),
+            account_summary(Context, cb_context:account_id(Context));
+        [{<<"websites">>,[]}] ->
+            lager:debug("getting system-wide summary"),
```
jamesaimonetti: What's the usecase for system-wide? Who would access those?

jamesaimonetti (changes requested):
Some quick notes to start

doc/websites.md
Navoda626: This was a initial parameter used in the doc but removed later. Will remove from the .md file also

priv/couchdb/schemas/websites.json
Navoda626: Will update this

doc/websites.md
Navoda626: Yes, will update this

doc/websites.md
Navoda626: will fix this

doc/websites.md
Navoda626: Will fix this

src/modules/cb_websites.erl
Navoda626: Yes this would not be used since admins also could see only their websites, will remove this

src/modules/cb_websites.erl
Navoda626: Currently it supports only binary types (ex: image/png), is it better to use base64 type as well?

doc/websites.md
Navoda626: Currently implemented
POST /{WEBSITE_ID}/logo
GET /{WEBSITE_ID}/logo
GET /{WEBSITE_ID} with accept header

So is it better to use only the following?
GET /{WEBSITE_ID} with accept header
POST /{WEBSITE_ID} with content-type header

icehess:
@Navoda626 Is this API has any use outside of commland app? We have a specific app in Kazoo for commland app api, `kazoo-desktop` which i think is better place to put these commland specific APIs

k-anderson:
@icehess howdy :) so while I understand that this API is only for the comm.land desktop application, there is a target to run comm.land as a website as well as mobile.  In both the alternate deployments comm.land would NOT require kazoo-desktop (that is providing access and APIs dedicated to the desktop deployments) so I do believe crossbar is the best place for this.  IE: when Monster UI is deprecated and comm.land is the browser UI, we will have environments that us comm.land without having the desktop deployment (payed option) so the external links will need to be used without kazoo-desktop present.

icehess:
@k-anderson thanks for clarification. That make sense now.

I know this is not a good place to dicuss this, but before I forget to bring it up, how do you think about a generic storage HTTP API so any app (commland, mui, kazoo) can call it to store their custom data, like a key-store storage? I'm not saying a fully fledged S3 API but also why not? :))) Though I understand it needs the scope and some other stuff to control its security and perhaps storage space quota. 

Like here instead of creating a specific API for anything that an UI app want to persistently store, we simply provide a common generic storage and commland-core can use it store. (I'm just using this website api as an example)

doc/websites.md
Navoda626: Implemented these API removing /logo APIs
GET /{WEBSITE_ID} with accept header
POST /{WEBSITE_ID} with content-type header

src/modules/cb_websites.erl
Navoda626: Base64 content-type is also implemented

PR #404, opened 2022-12-26
SUPP-217: Interaction pagination was missing the very first record of subsequent pages

ixmrm01:
When reduce and group level is involved, key length returned by CouchDB
can be shorter than actual key length of the view. Because of how
view collation works, this would cause one-off error in which the row with
those keys won't matched and not included in the result.

This adds the key_min_length options so kz_couch_view will correct start/end
keys accordingly.


src/modules/cb_cdrs.erl
```diff
@@ -307,8 +307,9 @@ get_view_options([{<<"cdrs">>, [?PATH_INTERACTION]}, {?KZ_ACCOUNTS_DB, _}|_]) ->
     ,props:filter_undefined(
        [{'range_start_keymap', []}
        ,{'range_end_keymap', crossbar_view:suffix_key_fun([kz_json:new()])}
+       ,{'key_min_length', 3}
        ,{'group', 'true'}
-       ,{'group_level', 2}
```
icehess, line 311: Why are you increasing the `group_level`. We use reduce to get a single record (the very first leg of the call) for the whole call. If you change this the reduce won't work as expected. We would end up with 2 records per call.

The real fix is just the line 310. No other changes is needed.

icehess (changes requested):

icehess:
closing in favor of https://github.com/2600hz/kazoo-crossbar/pull/414

PR #408, opened 2023-01-04
KCRO-122: Screenpop endpoint to segment screenpops per user

Navoda626:
KCRO-122: Screenpop endpoint to segment screenpops per user

Previously there was one default screenpop per account and all the comm.land users of the account could only use that default screenpop.

Now, the screenpops can be segmented per user using the schema change (new 'permissions' object in the document schema) and a new GET accounts/{ACCOUNT_ID}/users/{USER_ID}/screenpops endpoint is implemented to get user allowed screenpops. There are 3 types of permissions for a screenpop, all_users(boolean), allow(array of user ids) and deny(array of user ids). For a specific user id if the (all_users=true OR user_id in allow array) AND (user_id NOT in deny array) that screenpop is allowed.


src/modules/cb_screenpops.erl
```diff
@@ -156,6 +173,41 @@ load_summary(Context) ->
                       ,Options
                       ).
 
+-spec user_summary(cb_context:context(), kz_term:api_ne_binary()) -> cb_context:context().
+user_summary(Context, UserId) ->
+    Options = ['include_docs'
+              ,{'key', kzd_screenpops:type()}
+              ],
+    case kz_datamgr:get_results(cb_context:account_id(Context), <<"maintenance/listing_by_type">>, Options) of
```
icehess: This is basic Read operation for crossbar, so please don't use `kz_datamgr` directly. Use `crossbar_view:find/4` just like `load_summary/1`. The `crossbar_view` is crossbar way which handles pagination, limit and etc.. suitable for API interface.

src/modules/cb_screenpops.erl
```diff
@@ -156,6 +173,41 @@ load_summary(Context) ->
                       ,Options
                       ).
 
+-spec user_summary(cb_context:context(), kz_term:api_ne_binary()) -> cb_context:context().
+user_summary(Context, UserId) ->
+    Options = ['include_docs'
```
icehess: So if you're setting `include_docs` why adding `permission` inside the [view result](https://github.com/2600hz/kazoo-core/pull/563/files#diff-ac0e3217fe1475244d7e9e3cd11e50f6f58f9209911b3a1d8481f4a92c66809dR374) then?

If you need to use the `include_docs` please consider removing `defaultValue.permissions = doc.permissions;` from view to avoid making the view big for no propose.

src/modules/cb_screenpops.erl
```diff
@@ -156,6 +173,41 @@ load_summary(Context) ->
...
+        {'ok', SPs} ->
+            FilteredSPs = lists:filter(fun(Screenpop) ->
+                                               AllowAll = kz_json:is_true([<<"doc">>, <<"permissions">>, <<"all_users">>], Screenpop, 'true'),
+                                               AllowList = kz_json:get_list_value([<<"doc">>, <<"permissions">>, <<"allow">>], Screenpop, []),
+                                               DenyList = kz_json:get_list_value([<<"doc">>, <<"permissions">>, <<"deny">>], Screenpop, []),
+                                               (AllowAll
+                                                orelse lists:member(UserId, AllowList))
+                                                   andalso not lists:member(UserId, DenyList)
+                                       end, SPs),
+            ModifiedSPs = [modify_screenpops_output(FilteredSP) || FilteredSP <- FilteredSPs],
```
icehess: Instead of all of this, if you use `crossbar_view` you can add `{'mapper', normalize_user_summary/2}` options (see `load_summary/1` too).

`crossbar_view` then automatically calls that functions and you can then filter and normalize the object however you want:

```
normalize_user_summary(JObj, Acc) ->
    AllowAll = kz_json:is_true([<<"doc">>, <<"permissions">>, <<"all_users">>], Screenpop, 'true'),
    AllowList = kz_json:get_list_value([<<"doc">>, <<"permissions">>, <<"allow">>], Screenpop, []),
    DenyList = kz_json:get_list_value([<<"doc">>, <<"permissions">>, <<"deny">>], Screenpop, []),
    case (AllowAll
        orelse lists:member(UserId, AllowList))
        andalso not lists:member(UserId, DenyList)
    of
        kz_json:get_value(<<"value">>, JObj);
....
....
....
```

icehess (changes requested):

src/modules/cb_screenpops.erl
Navoda626: Okay. will change that to crossbar_view:find/4

src/modules/cb_screenpops.erl
Navoda626: modified this to use crossbar_view:find/4, so not using 'include_docs' now. So the change in the view is not changed

src/modules/cb_screenpops.erl
Navoda626: This is changed as a normalize_user_summary

src/modules/cb_screenpops.erl
```diff
@@ -156,6 +173,30 @@ load_summary(Context) ->
                       ,Options
                       ).
 
+-spec user_summary(cb_context:context(), kz_term:api_ne_binary()) -> cb_context:context().
+user_summary(Context, UserId) ->
+    Selector = [{'start', [{<<"doc_type">>, kzd_screenpops:type()}]}
+               ,{'end', [{<<"doc_type">>, kzd_screenpops:type()}]}
+               ],
+    Options = [{'mapper', fun (JObjs) -> normalize_user_summary(UserId, JObjs) end}
```
icehess, line 181: hmm, I'm not very comfortable with getting a potentially huge list of user permissions and then just looking for `true` or a single user id. Also that is why I don't like to include `permissions` in the `value` of the `crossbar_listings/by_type_id` view for screenpops.

Any idea how big this permission could be potentially become? like lets say we have 1000 users in the account, and you create a screenpops for 500 users. With this approach here:

1. the couchdb `crossbar_listings` view index file grows significantly, this reduce the speed of couchdb view and increases the disk space usage
2. to list screenpops for a single user, you are fetching a huge doc (with all those 500 user ids) and then only checking for a user id that may not even be in that list, this waste memory space and time and pressures more the db and kazoo crossbar

Always try put things on scale :)) and recognize the bottle neck. The thing i said above maybe not even happen, depends on how people use screenpops, but it still there and could cause the problem in the future.


Another approach is to create a a dedicated view just only for screenpops, and add a view that emits the a single row per `doc.permissions`. Then you can simply use this Couchdb view option to query:
```
{'keys', ['true', UserId]}
```

icehess (changes requested):

src/modules/cb_screenpops.erl
Navoda626: I agree if a separate view is defined this would be better. But at the initial discussion it was said there will be around 5 screepops per account and so that to go with the application filter rather than use map reduce function for the database.

icehess: well i don't know anything screenpop, I was just expressing my concern here as you have more knowledge about this screenpops and its usage.

priv/couchdb/schemas/screenpops.json
```diff
@@ -59,6 +59,33 @@
...
+            "properties": {
+                "all_users": {
+                    "default": true,
+                    "description": "Whether the screenpop allowed for all users in the account",
+                    "type": "boolean"
+                },
+                "allow": {
+                    "default": [],
+                    "description": "Specific list of users who are allowed the screenpop",
+                    "items": {
```
icehess, line 74: Per our discussion, please set maximum array size (50):

```suggestion
                    "maxItems": 50,
                    "items": {
```

priv/couchdb/schemas/screenpops.json
```diff
@@ -59,6 +59,33 @@
...
+                        "type": "string"
+                    },
+                    "type": "array"
+                },
+                "deny": {
+                    "default": [],
+                    "description": "Specific list of users who are not allowed the screenpop",
+                    "items": {
```
icehess, line 83: Per our discussion, please set maximum array size (50):

```suggestion
                    "maxItems": 50,
                    "items": {
```

src/modules/cb_screenpops.erl
icehess: Okay, we discuss this in engineer town hall and we decided that we don't expect people set too many user ids in the array.

But also as per caution to avoid bad configuration we decide to set a maximum length for the arrays. The decide max number is 50.

 I made the suggestion code above.

icehess (changes requested):

priv/couchdb/schemas/screenpops.json
Navoda626: This is done

priv/couchdb/schemas/screenpops.json
Navoda626: This is done

doc/screenpops.md
```diff
@@ -26,13 +26,19 @@ Key | Description | Type | Default | Required | Support Level
...
+`permissions.deny` | Specific list of users who are not allowed the screenpop | `array(string())` | `[]` | `false` |  
+`permissions` |   | `object()` | `{}` | `false` |  
 `post_call_extension_time` | Contains the time the screenpop will remain active when the temporal type is selected. | `number()` |   | `false` |  
 `show_location_call_time` | Flag field for location and local time of customer call | `boolean()` | `false` | `false` |  
 `type_screenpops` | Type of screenpops | `string()` | `notification` | `false` |  
 
 
 
-#### Fetch
+#### Fetch all screenpops for an Account
```
jamesaimonetti, line 43: Doc updates look great! My only ask is to add a section about the `permissions` object and state explicitly what the precedence is.

Something like, "KAZOO first considers if `all_users=true`; if so, permissions checks stop. If false, the deny list is first checked to bar the specified user explicitly. If the user isn't present in the deny list, check if the user is explicitly listed in the allow list. If not in the allow list, the user is barred."

Don't copy/paste that! But hopefully it gets the idea across; guide the client using this field on what should occur given different combinations of all_users and allow/deny lists.

src/modules/cb_screenpops.erl
```diff
@@ -156,6 +173,30 @@ load_summary(Context) ->
...
+              ],
+    crossbar_view:find(Context
+                      ,<<"crossbar_listings/by_type_id">>
+                      ,Selector
+                      ,Options
+                      ).
+
+normalize_user_summary(UserId, SPs) ->
+    lists:filter(fun(Screenpop) ->
+                         AllowAll = kz_json:is_true([<<"value">>, <<"permissions">>, <<"all_users">>], Screenpop, 'true'),
```
jamesaimonetti: I'd like to see the "value" pulled out and then use kzd_screeenpops:permissions_* functions to access the all_users/allow/deny values.

doc/screenpops.md
Navoda626: Will update the logic behind permissions

src/modules/cb_screenpops.erl
Navoda626: This is updated

src/modules/cb_screenpops.erl
Navoda626: Done setting maximum lengths

PR #414, opened 2023-01-25
[master] KCRO-172: fix interaction listing one-off issue by using key_min_length options

icehess:
When reduce and group level is involved, key length returned by CouchDB can be shorter than actual key length of the view. Because of how view collation works, this would cause one-off error in which the row with those keys won't matched and not included in the result.

This adds the key_min_length options so kz_couch_view will correct start/end keys accordingly.


icehess:
5.1 PR: https://github.com/2600hz/kazoo-crossbar/pull/415
properly test: https://github.com/2600hz/kazoo-properly/pull/121

jamesaimonetti:
@icehess do we have an example of what returned before this change, and what after, that could be used in properly?

icehess:
@jamesaimonetti I was working on properly for test this, but got distracted by other things, will get back to it and create/link PR here later

icehess:
properly test: https://github.com/2600hz/kazoo-properly/pull/121

icehess:
thanks @jamesaimonetti, test was already there for owner, I just needed to copy paste and remove the owner :)

PR #417, opened 2023-02-07
KZOO-350: Add case to handle undefined PresenceId

dhuynh-Kazoo:


src/modules/cb_presence.erl
```diff
@@ -369,6 +369,7 @@ publish_presence_reset(Context, PresenceId) ->
     kz_amqp_worker:cast(API, fun kapi_presence:publish_reset/1).
 
 -spec publish_presence_update(cb_context:context(), kz_term:ne_binary(), kz_term:ne_binary()) -> 'ok'.
```
lazedo:
```suggestion
-spec publish_presence_update(cb_context:context(), kz_term:api_ne_binary(), kz_term:ne_binary()) -> 'ok'.
```

lazedo (changes requested):

src/modules/cb_presence.erl
dhuynh-Kazoo: Hi @lazedo, the type for 2nd argument is updated

lazedo (approved):

PR #421, opened 2023-02-15
KCRO-141: Match lists

jamesaimonetti:
Match lists allow the API developer to specify a list of rules to evaluate to determine if a condition is met successfully.

Match lists can be used in a number of situations:

1. Callflow branching
2. Caller ID selection
3. Call Forwarding selection
4. Block lists

Match lists are a generic form of encoding that condition checking; the actual execution of the check will depend on the context it is run in - for instance callflows will evaluate the match list rules against the call record.


jamesaimonetti:
core PR: https://github.com/2600hz/kazoo-core/pull/583
properly for CRUD: https://github.com/2600hz/kazoo-properly/pull/125

PR #422, opened 2023-02-16
KCRO-168: Validating the getstream feature enabled in user document

Navoda626:
KCRO-168: Validating the getstream feature enabled in user document

Earlier when the getstream API PUT /accounts/{ACCOUNT_ID}/users/{USER_ID}/getstream/connect is called user is connected to getstream API without validating whether the user has getstream feature (chat.getstream.emabled=true).

Now, added chat.getstream.emabled parameter is the user document schema and when getstream API calls at cb_getstream.erl validate whether the user document parameter chat.getstream.emabled=true or otherwise return a 400 disabled response


nkarunarathne:
Changes look good

src/modules/cb_getstream.erl
```diff
@@ -69,15 +69,25 @@ content_types_provided(Context, _) ->
...
--spec validate_request(cb_context:context(), path_token(), list(), req_verb()) -> cb_context:context().
-validate_request(Context, ?CONNECT_PATH, UserId, ?HTTP_PUT) ->
+-spec validate_user_feature(cb_context:context(), path_token(), list(), req_verb()) -> cb_context:context().
+validate_user_feature(Context, ?CONNECT_PATH, UserId, ?HTTP_PUT) ->
     Context1 = crossbar_doc:load(UserId, Context),
-    case cb_context:resp_status(Context1) of
-        'success' -> Context2 = validate_secret(Context1),
-                     validate_api_key(Context2);
-        _Else -> Context1
+    Enabled = kz_json:get_value([<<"chat">>, <<"getstream">>, <<"enabled">>], cb_context:doc(Context1), 'false'),
```
jamesaimonetti: You added the nice accessor functions, let's use them here

Navoda626: Added the accessor function

Navoda626:
Required https://github.com/2600hz/kazoo-core/pull/586

jamesaimonetti:
@Navoda626 discussing this further, the recommendation is to manage getstream configs as a pvt field on the account/user doc (depending on the URI) to manage the config. Let's talk more on a training call but wanted to put the idea in your head.

PR #424, opened 2023-02-17
KZOO-270: Migrate docs with legacy call_forward settings

jamesaimonetti:
If legacy enabled=true, set new match_list.enabled = true If legacy enabled=false, set new match_list.enabled = false


jamesaimonetti:
Test migrate with local test: https://github.com/2600hz/kazoo-properly/pull/126
Core PR for accessors: https://github.com/2600hz/kazoo-core/pull/592

PR #429, opened 2023-02-23
KCRO-173: Add an optional field open_in_browser to the website document

Navoda626:
KCRO-173: Add an optional field open_in_browser to the website document

Adding a boolean optional parameter 'open_in_browser' default set as 'false' for the website document so that the admin or the user adding the website can specify that url to be opened specifically in browser other than the commland application (Since some websites doesn't open as expected in commland application, so those need to be opened in browser).


sanjeevie6353:
@Navoda626 can you please correct this small typo...?  :)      an website -> a website

`Whether an website should be opened in browser`

The CircleCI fails are only due to dependency with https://github.com/2600hz/kazoo-core/pull/596 PR.

Navoda626:
@sanjeevie6353 It's fixed. Thanks

Navoda626:
Required PR: https://github.com/2600hz/kazoo-core/pull/596

PR #435, opened 2023-03-08
[5.2] KCRO-173: Add an optional field open_in_browser to the website document

Navoda626:
Adding a boolean optional parameter 'open_in_browser' default set as 'false' for the website document so that the admin or the user adding the website can specify that url to be opened specifically in browser other than the commland application (Since some websites doesn't open as expected in commland application, so those need to be opened in browser).

(cherry picked from commit 9bc16ce2cce301e0d41178cb231890df72076ff6)


Navoda626:
Required https://github.com/2600hz/kazoo-core/pull/612

PR #437, opened 2023-03-10
Patch resources without losing nested keys

jamesaimonetti:
It was observed that PATCH-ing a resource with a nested key path (like classifiers.unknown.enabled) would remove other key paths in the same object.

The change is to use the standard crossbar_doc:patch_and_validate/3 instead of the custom and wrong merging of objects.


jamesaimonetti:
Test with bug reproduced and confirming the fix: https://github.com/2600hz/kazoo-properly/pull/132

harenson (approved):
👍🏾

PR #445, opened 2023-03-14
KZOO-308 Functionality to Intercept Users According to CALEA Intercept Requests

nkarunarathne:
Added configuration definitions of wiretap-related configurations


lazedo:
this is a private application and `{schemas_to_priv, true}`  is set in the app. this shouldn't be here.

PR #446, opened 2023-03-16
KZOO-309: Extend addresses functionality to users and accounts

harenson:
Depends on https://github.com/2600hz/kazoo-core/pull/619
Tests https://github.com/2600hz/kazoo-properly/pull/133

- Extend addresses functionality to users and accounts. It was in devices only.

- Update _emergency address_ to have more meaningful field names and also to include some optional fields handled by PIDF+LO.

- Also, since users used to have an existing, non-documented, _addresses_ field to build the user's vcard, a migrate mechanism was added to migrate it into the new format/schema.


priv/couchdb/schemas/callflows.resources.json
```diff
@@ -3,6 +3,15 @@
     "_id": "callflows.resources",
     "description": "Validator for the resources callflow data object",
     "properties": {
+        "addresses": {
+            "properties": {
+                "emergency": {
+                    "description": "",
```
sanjeevie6353: Is it ok to have this parameter as `""`?

icehess: Documenting things is encouraged!

src/crossbar_doc.erl
```diff
@@ -238,7 +238,38 @@ handle_successful_load(Context, JObj, 'false', _) ->
...
+%% This is being done at "load" step, to make sure the returned document has the
+%% new addresses format, in case that document is used later in a POST request.
+%% Also, it allows the patch_and_validate/4 function to avoid overriding the
+%% existing addresses value in case the user already has an old format addresses
+%% value, and the request is to PATCH the user with an emergency address. In
+%% this scenario, the user should end up with 2 addresses: emergency and vcard.
+%% @end
+%%------------------------------------------------------------------------------
+-spec maybe_migrate_user_addresses(kz_json:object()) -> kz_json:object().
+maybe_migrate_user_addresses(JObj) ->
```
icehess, line 262: If the concerns is only regarding POST/PUT/PATCH and only for `user` doc then why not move this migration to `kzd_users:validate/3`?

This seems off to do this here and only for a specific doc type. This opens up this practice to do these one off things this way and then we end up with some spaghetti code, thought it is delicious but is a mess to cleanup.

Or if you really need to also make sure we have proper new format only for GET (like UI is expecting this) then the suggestion would be to not break things. But since that ship has sailed, another place to do this migration would be writing a sup command to do the migration. and add change note that need release need that migration command.

icehess: Or at least do it in `cb_users`?

harenson: @icehess Doing it within cb_users requires a lot more work. While doing it here, it will be automatically handled, not matter what HTTP verb is being used. GET and POST are easy, but PATCH is the one that requires more work. It would require this migration process to _maybe_ happen for different Context's fields (db_doc, req_data, resp_data, and I don't remember what others) and those changes cannot all be made at the same time.

icehess: You can some new function in `crossbar_doc` for this:

```erlang
  -spec load_merge(kz_term:ne_binary(), kz_json:object(), cb_context:context(), kz_term:proplist(), kz_term:api_object()) ->
            cb_context:context().
  load_merge(DocId, DataJObj, Context, Options, 'undefined') ->
      Context1 = load(DocId, Context, Options),
      merge_loaded(DataJObj, Context);
  load_merge(_DocId, _DataJObj, Context, _Options, BypassJObj) ->
      handle_datamgr_success(BypassJObj, Context).

  -spec merge_loaded(kz_json:object(), cb_context:context()) -> cb_context:context().
  merge_loaded(DataJObj, Context) ->
      case 'success' =:= cb_context:resp_status(Context) of
          'false' -> Context;
          'true' ->
              lager:debug("loaded doc ~s(~s), merging", [DocId, kz_doc:revision(cb_context:doc(Context))]),
              Merged = kz_json:merge_jobjs(kz_doc:private_fields(cb_context:doc(Context)), DataJObj),
              handle_datamgr_success(Merged, Context)
      end.
```

PR #448, opened 2023-03-17
KZOO-338: Commland quick-call sometime fail to proceed

Navoda626:
Previously, when using the quick-call function in commland there were 2 ecallmgr_call_control processes started and sometimes with a 'control usurp' event both the control processes usurped resulting in end of quick-call. Here control process start with the originate request for the ecallmgr from quick-call API is not necessary.

Now, setting Start-Control-Process header to false in the originate request for the ecallmgr in the quick-call scenario so that the unwanted call control process will not be started and the call proceeds as expected.


lazedo (changes requested):
you are changing behaviour because of the usage from one application. you need to provide a way to retain current behaviour and allow new behaviour. like i mentioned before, maybe the problem is the way that commland is using this, maybe they should be using the conference api to dial out to user endpoint or number

Navoda626:
Agreed on this. Closing the PR.

PR #453, opened 2023-03-27
KCRO-176 Authentication should be prioritised when account details are queried through the API.

nkarunarathne:
Previously, when fetching account details through API, account id validation was prioritized over token authentication.

Now token authentication is prioritized over account id validation. Therefore when auth token is invalid system returns HTTP 401 instead of HTTP 404.


src/api_resource.erl
```diff
@@ -493,7 +493,16 @@ malformed_request(Req, Context) ->
           {boolean(), cowboy_req:req(), cb_context:context()}.
 malformed_request(Req, Context, ?HTTP_OPTIONS) ->
     {'false', Req, Context};
-malformed_request(Req, Context, _ReqVerb) ->
+malformed_request(Req, Context, ReqVerb) ->
+    malformed_request(Req
+                     ,Context
+                     ,ReqVerb
+                     ,validate_auth_token(cb_context:auth_token(Context))).
```
icehess: As we discussed the crossbar rest api is as follow:

```
init/rest_init (parses the path and creates cb_context)
-> service_available 
-> known_methods
-> allowed_methods
-> early_authenticate (called in `api_resource:allowed_methods`)
-> malformed_request (checks the account is correct)
-> is_authorize (calls authenticate bindings)
-> forbbiden (calls authorize bindings)
```

as you can see we have two steps for authentication:
1. early_authenticate: this performs the authentication way before `is_authorize` is called by cowboy_rest behavior. I don't recall why we do this early auth, but this how it is now. Currently `cb_token_auth` is the only auth modules that is binding to this. ANd it only works withx-auth-token if it is PRESENT in the request header. It doesn't work with basic auth and etc... So this function may return `true` for early_auth. This means we are not done yet fully
2. Now in `is_authorize` we do another check with running authenticate bindings to ask all auth modules. THis is only happens if the auth_doc doesn't exists yet. That means early auth couldn't find the proper header and blindly returned true. So this is the really auth and calls all auth modules this time if it fails is is a hard stop.

Here you are only returning invalid_cred error only IF the x-auth-token present and it is not valid. But it still let through the other auth mechanism and/or if x-auth-token wasn't in header at all.

So this is tricky. My previous approach was to move this account id check to is_authorized step. But I don't think this evens works. There are a couple of auth modules that only check for path and return true. So even if authenticate return true, this doesn't mean that we can return the actual bad account id yet. The attacker can forge a request to use a URL path that they know would pass authenticate step and still can check if the account id exists or not.


I think the best solution  here to change the case clause below and always return invalid_credentials if the account id is invalid. We lose ability to tell if account id correct or not for legitimate request, but still this a good comprise for a better security.

Also I suggest to bring this up in engineering channel to get the idea of other erlang engineers.

icehess (changes requested):

src/api_resource.erl
icehess: In other words, you're approach doesn't work because  `malformed_request` because `early_authentic` only checks for `x-auth-token` header. So the only thing that the attacker need to do is to not send the `x-auth-token` and all is good again and it get `bad_identifier`. 

We cannot do it after `is_authorize` step either. Why? For example an attacker can forge this URL:

GET or POST to`/v2/accounts/fake_account_id/clicktocall/{C2CId}/connect`

So now `cb_clicktocall:authenticate/3` will respond to the binding of `authenticate` and returns `true` which means the request is authenticated and allows to advance. We now moved the account id check after we got `true` for auth and auth was successful so now we return bad_identifier again.

One thing you can check is to search for any `cb_` file that implemented `authenticate` callback function and follow what their are doing. Most of them are okay and strictly check that URL doesn't have account id. But something like clicktocall here are problematic. Also cb_whitelable, cb_tasks, cb_devices, cb_users.

```
grep -E '^authenticate\(' applications/crossbar/src/modules/*
```

jamesaimonetti: Added a check for the c2c example: https://github.com/2600hz/kazoo-properly/pull/135/commits/82e40fa94f3408b3aa33930d2b7b26b2fb83ae9e

icehess:
CI is failing because of `seq_phone_number` test checks for bad identifier when account is invalid. This PR is fixing that issue. So the properly is updating the test to expected 401 error.

Also that properly test checks for basic auth, x-auth-token, no x-auth-token, invalid x-auth-token and valid x-auth-token to make sure we always return invalid credentials.

Require PR:  https://github.com/2600hz/kazoo-properly/pull/135

PR #461, opened 2023-04-03
ensure proper mapper fun in crossbar view

lazedo:
* detected by chance on a misused case of crossbar view wher {mapper, fun crossbar_view:get_value_fun/0}  was passed instead of {mapper, crossbar_view:get_value_fun()}


src/crossbar_view.erl
```diff
@@ -401,6 +402,32 @@ build_cursor_options(Context, Options, NoDesignModule, IsTimeQuery) ->
...
+        undefined -> get_value_fun();
+        Fun -> verify_mapper_fun(Fun)
+    end.
+
+-spec verify_mapper_fun(term()) -> user_mapper_fun().
+verify_mapper_fun(Fun) when is_function(Fun, 1) -> Fun;
+verify_mapper_fun(Fun) when is_function(Fun, 2) -> Fun;
+verify_mapper_fun(Fun) when is_function(Fun, 3) -> Fun;
+verify_mapper_fun(Fun) when is_function(Fun, 0) ->
+    case Fun() of
```
jamesaimonetti: Do we want to support this? I would either do `lager:error` so CI could fail if this happens, or crash here, to be loud to the developer that something is amiss.

lazedo: sure, was just to trying to address an issue with already deployed app

icehess:
oh sorry my bad, those `get_*` function was shortcut to not use `fun` syntax.

PR #463, opened 2023-04-06
KCRO-181: ignore empty mms body while there is a at least a file

icehess:
This will remove minimum from MMS schema body field and improves the cb_mms to make sure we ignore empty body ONLY when there is at least a file in the request.

If no file and empty body then the request fails as before with validation error.


icehess:
PRs:
- master:
    - https://github.com/2600hz/kazoo-core/pull/645
    - https://github.com/2600hz/kazoo-crossbar/pull/463

icehess:
CI is failing because of circular pr dependency on https://github.com/2600hz/kazoo-core/pull/645

PR #466, opened 2023-04-14
KZOO-376: treat /v2/accounts as /v2/accounts/{AUTH_ACCOUNT_ID}

jamesaimonetti:


jamesaimonetti:
Tests at https://github.com/2600hz/kazoo-properly/pull/144
Core changes needed at https://github.com/2600hz/kazoo-core/pull/649

k-anderson:
Idea: we have talked about a /my/ endpoint, thoughts on /account/user/ (dropping the plural) rather than '/my/' or this syntactical sugar?

jamesaimonetti:
> Idea: we have talked about a /my/ endpoint, thoughts on /account/user/ (dropping the plural) rather than '/my/' or this syntactical sugar?

https://github.com/2600hz/kazoo-crossbar/blob/master/src/api_util.erl#L805

We treat `/v2/account` as `/v2/accounts/{AUTH_ACCOUNT_ID}`

PR #467, opened 2023-04-17
KZOO-301: Add a period of user lockout for failed login attempts

dhuynh-Kazoo:
Add simple account lock out mechanism using Token Buckets if any of the account's user failed to login with correct creds  in a configurable attempt tries.


src/cb_modules_util.erl
```diff
@@ -163,6 +163,10 @@ parse_media_type(MediaType) ->
     end.
 
 -spec bucket_name(cb_context:context()) -> kz_term:ne_binary().
+bucket_name({account_id, undefined}) -> <<"no_account">>;
+bucket_name({account_id, AccountID}) -> <<AccountID/binary>>;
+bucket_name({ip, undefined}) -> <<"no_ip">>;
+bucket_name({ip, IP}) -> <<IP/binary>>;
```
icehess: Revert this, you are changing bucket name for everything. We don't want to do that.

icehess (changes requested):
My assumption was that only crossbar auth modules were consuming token buckets. But after meeting with Erlang team I noticed I was wrong (well kind, but it is complicated and unrelated :))).

So instead, I suggest we add new set of functions to `cb_modules_util` for dealing with auth modules specifically.

1. `cb_modules_util:auth_bucket_names(Context, BucketNames)`:
    - This checks `Context` and adds `client_ip` (if any) to the list of passed `BucketNames`.
    - Each auth module must call this, and they have to pass a list of BucketNames.
    - These BucketNames are specific to each auth module, and depends on what they accept as payload.
    - Check each auth modules `validate` function and follow the code to see if they are expecting any payload and if yes what is the payload.
    - As bare minimum we are interested in account_id, account_name, account_realm
2. `cb_modules_util:consume_auth_tokens_until(Context, TokenCost, BucketNames)`:
    - All cb auth modules must call this instead of normal function.
    - This should iterate over bucket names and call `kz_buckets:consume_tokens_until/3` and immediately stops if it returns `false`
    - In each iteration track the bucketname and 
    - Instead of using `?APP_NAME` as first argument for `kz_buckets:consume_tokens_until/3`, use `crossbar_authenticate`. This allows us to configure how fast enough we want to fill the bucket for authentications.
    - Regardless of what `kz_buckets:consume_tokens_until/3` response is for auth buckets, also call regular `cb_modules_util:consume_tokens_until(Context, TokenCost)` to count against regular token bucket of `crossbar` so we penalty the client for the rest of endpoints too. This also set remaining token cost when the crossbar builds the response envelope.
    - the return value is is the new context updated by regular `consume_tokens_until` and a boolean indicating any of these consume tokens were false.
3. If `cb_modules_util:consume_auth_tokens_until(Context, TokenCost, BucketNames)` returns `{false, NewContext}`. stop the request and add system error, something like: `cb_context:add_system_error(401, 'invalid_credentials', <<"too many failed login attempt, your account is locked">>, Context)`
4. There were some thoughts on adding a private field to the account doc to indicate the account is locked, or other mechanism to sync all nodes on this lockout. I don't have any suggestion on this, feel free to bring this up with the rest of erlang engineers. 
5. List of auth modules by my count:
    - cb_desktop_auth (from kazoo-desktop repo)
    - cb_user_auth
    - cb_api_auth
    - cb_basic_auth
    - cb_conference_auth

Not sure @jamesaimonetti wants to add or change anything?

dhuynh-Kazoo:
hi @icehess , follow your suggestion for the function `cb_modules_util:auth_bucket_names(Context, BucketNames)`. I made some changes. Here is my idea about that:

I think each module has one bucket( I checked your comment about multiple buckets for one module and compare them in the JSON file but I am not so sure that we can have multiple buckets).

 And in my implementation, the bucket name format is `<client_ip>/<auth_module>/<client_id>` or <client_ip>/<auth_module>. The <client_id> is got from the `cb_modules_util:validate_auth_parameter`, the function may return `undefined` if client_id/account_name/account_realm are undefined, otherwise it can return any valid value. 

So at this point, we can make sure each auth module has one bucket, and based on the above Context fields, we will have the corresponding name. Please check the commit, if you have any ideas about multiple buckets, please clear it again.

dhuynh-Kazoo:
Hi @icehess, 
From your 2nd suggestion `Instead of using ?APP_NAME as first argument for kz_buckets:consume_tokens_until/3, use crossbar_authenticate. This allows us to configure how fast enough we want to fill the bucket for authentications.`. Is the `crossbar_authenticate` a new config parameter then we will add? I tried to find this keyword but I could not find it in any places in the code base.

src/cb_modules_util.erl
```diff
@@ -162,6 +163,14 @@ parse_media_type(MediaType) ->
             {'error', 'badarg'}
     end.
 
+-spec auth_bucket_names(cb_context:context(), string()) -> kz_term:ne_binary().
+auth_bucket_names(Context, AuthMod) ->
+  case validate_auth_parameter(Context) of
+    'undefined' -> bucket_name(cb_context:client_ip(Context), AuthMod);
+    ValidParam ->
+        bucket_name(cb_context:client_ip(Context), <<AuthMod/binary, "/", ValidParam/binary>>)
```
icehess: We want different bucket name for different parameters. creating a single bucket name consist of client ip and something else potentially like account id doesn't change anything regarding what are already doing.

Instead of creating a single bucket, return a list of bucket names:

`[client_ip, some_form_of_account_identity_like_id_realm_name_or_phone_number]`

src/cb_modules_util.erl
```diff
@@ -492,3 +501,15 @@ convert_numbers_to_list(Key, Value, JObj) ->
     [kz_json:from_list([{<<"number">>, Key} | kz_json:recursive_to_proplist(Value)])
     | JObj
     ].
+
+-spec validate_auth_parameter(cb_context:context()) -> 'undefined' | kz_term:api_ne_binary().
+validate_auth_parameter(Context) ->
+  Params = [Fun(Context) || Fun <-  [fun cb_context:account_id/1
+                                   ,fun cb_context:account_name/1
+                                   ,fun cb_context:account_realm/1
```
icehess: As we discuss this before, these `cb_context:account_id` and such are using the account id from URL. If you see any URL when doing authenticate there is no account id in the URL, and these always return undefined defeating the propose of what we are doing here.

```
For example for user auth
PUT /v2/user_auth

so cb_context:account_id/1 always returns undefined
```

Instead of this you have to look inside the payload
```
```suggestion
    ReqData = cb_context:req_data(Context),
    Params = [Fun(Context) || Key <-  [<<"account_id">>
                                      ,<<"account_name">>
                                      ,<<"account_realm">>
                                      ,<<"phone_number">>
                                      ],
                              Value <- [kz_json:get_ne_binary_value(Key, ReqData)],
                              Value =/= 'undefined'
             ]
```

src/modules/cb_api_auth.erl
```diff
@@ -98,7 +98,8 @@ authenticate_nouns(_Nouns) -> 'false'.
 -spec validate(cb_context:context()) -> cb_context:context().
 validate(Context) ->
     TokenCost = cb_modules_util:token_cost(Context, ?API_AUTH_TOKENS),
-    case cb_modules_util:consume_tokens_until(Context, TokenCost) of
+    BucketName = cb_modules_util:auth_bucket_names(Context, ?MODULE, cb_context:account_name(Context)),
+    case cb_modules_util:consume_tokens_until(Context, TokenCost, BucketName) of
         {'true', Context1} ->
```
icehess: Here after this line we want also to call `cb_modules:consume_auth_tokens_until/1` to check for user lockout.
```suggestion
    case cb_modules_util:consume_tokens_until(Context, TokenCost, BucketName) of
        {'true', Context1} ->
            AuthBucketNames = cb_modules_util:auth_bucket_names(Context1, []),
            case cb_modules_util:consume_auth_tokens_until(Context1, AuthBucketNames, Cost) of
                'true' ->
```

icehess (changes requested):

src/modules/cb_api_auth.erl
icehess: And then in `cb_modules_util:consume_auth_tokens_until(Context, AuthBucketNames, Cost)` will do something like this by fold over BUcketNames and calls:

```erlang
kz_buckets:consume_tokens_until(<<"crossbar_auth">>, BucketName, Cost)
```

src/cb_modules_util.erl
dhuynh-Kazoo: Thank you for your suggestion but I imagine that for a bucket named `client_id`, we may used this bucket for all other auth modules like `cb_api_auth.erl, cb_conference_auth.erl and cb_user_auth.erl`. Is it a problem because we want to have buckets separately for each module?

dhuynh-Kazoo: Hi @icehess, Could you please explain the purpose that we should have multiple buckets based on each auth module? For example with cb_user_auth.erl, we only have account_name in the request payload. If the user tries with multiple account_name, the cb_modules_util:auth_bucket_names(Context, BucketNames)` may create multiple buckets. Is it worth it?

My idea is that we should have a generic bucket for all modules (we already created one from cb_modules_util:bucket_names) and a new one named <client_ip>/<auth_module> for each type of auth. Here is my draft implementation, 

First, we check if the user's request still has the remaining token in generic bucket, then we continue to check the token bucket for the current auth module 

```
consume_tokens(Context) ->
    TokenCost = cb_modules_util:token_cost(Context, ?USER_AUTH_TOKENS),
    %%  mutual bucket: created from bucket_name fun, used between modules
    %%  Auth Module: specific for each auth module
    AuthBucket = cb_modules_util:auth_bucket_name(Context, kz_term:to_binary(?MODULE)),

    case cb_modules_util:consume_tokens_until(Context, TokenCost) of
        {'true', Context1} ->
          case cb_modules_util:consume_tokens_until(Context1, TokenCost, AuthBucket) of
            {'true', Context2} -> {'ok', Context2};
            {'false', Context2} ->
              cb_context:add_system_error(401,
                'invalid_credentials',
                <<"too many failed login attempts, your account is locked">>,
                Context2)
          end;
      {'false', Context1} -> {'error', cb_context:add_system_error('too_many_requests', Context1)}
    end.

```

icehess: I never said we need separate bucket names for each auth module. we want ALL auth module use parameters that exists on their request data like account id, name, realm or phonenumber as a list of bucket names to consume.

dhuynh-Kazoo: Got it. Should we tie the client id to user parameters or just have the user parameter as the bucket name?

src/cb_modules_util.erl
```diff
@@ -162,6 +163,23 @@ parse_media_type(MediaType) ->
             {'error', 'badarg'}
     end.
 
+-spec auth_bucket_names(cb_context:context()) -> list().
+auth_bucket_names(Context) ->
+  [cb_context:client_ip(Context) | valid_user_identities(Context)].
```
icehess: On second thought we don't even need client ip, remove it. move valid user identities here.

src/cb_modules_util.erl
```diff
@@ -213,6 +231,14 @@ consume_tokens(Context, BucketName, Cost, ConsumeFun) ->
     ,cb_context:setters(Context, Setters)
     }.
 
+consume_auth_tokens(Context, BucketName, Cost) ->
```
icehess: we don't need this function remove completely.

src/cb_modules_util.erl
```diff
@@ -492,3 +518,14 @@ convert_numbers_to_list(Key, Value, JObj) ->
...
+valid_user_identities(Context) ->
+    ReqData = cb_context:req_data(Context),
+    [Value || Key <-  [<<"account_id">>
+                      ,<<"account_name">>
+                      ,<<"account_realm">>
+                      ,<<"phone_number">>
+                      ]
+                      ,Value <- [kz_json:get_ne_binary_value(Key, ReqData)]
+                      ,Value =/= 'undefined'
+    ].
```
icehess: Move this to `auth_bucket_names/1`

src/cb_modules_util.erl
```diff
@@ -162,6 +163,23 @@ parse_media_type(MediaType) ->
             {'error', 'badarg'}
     end.
 
+-spec auth_bucket_names(cb_context:context()) -> list().
+auth_bucket_names(Context) ->
+  [cb_context:client_ip(Context) | valid_user_identities(Context)].
+
+-spec consume_auth_tokens_until(cb_context:context(), non_neg_integer(), list()) ->
+          {boolean(), cb_context:context()}.
+consume_auth_tokens_until(Context, TokenCost, []) ->
```
icehess: Remove this and use replace it with this:
```erlang
-spec consume_auth_tokens_until(non_neg_integer(), list()) ->
          {boolean(), cb_context:context()}.
consume_auth_tokens_until(_TokenCost, []) ->
    'true';
consume_auth_tokens_until(TokenCost, [BucketName|BucketNames]) ->
    case kz_buckets:consume_tokens_until(<<"crossbar_authenticate">>, BucketName, Cost) of
      'true' -> consume_auth_tokens_until(TokenCost, BucketNames);
      'false -> 'false
    end.
```

src/modules/cb_api_auth.erl
```diff
@@ -98,7 +98,8 @@ authenticate_nouns(_Nouns) -> 'false'.
 -spec validate(cb_context:context()) -> cb_context:context().
 validate(Context) ->
     TokenCost = cb_modules_util:token_cost(Context, ?API_AUTH_TOKENS),
-    case cb_modules_util:consume_tokens_until(Context, TokenCost) of
+    BucketName = cb_modules_util:auth_bucket_names(Context, kz_term:to_binary(?MODULE)),
```
icehess: First call regular consume_token_until (which will do it job and check bucket for `client_ip/no_account_id`.
The call if true call consume_auth_tokens_until.

as i showed before something like this:

```erlang
    case cb_modules_util:consume_tokens_until(Context, TokenCost, BucketName) of
        {'true', Context1} ->
            AuthBucketNames = cb_modules_util:auth_bucket_names(Context1, []),
            case cb_modules_util:consume_auth_tokens_until(Context1, AuthBucketNames, Cost) of
                'true' -> ....
                'false' ->
                    {'error', cb_context:add_system_error(401, 'invalid_credentials', <<"too many failed login attempt, your account is locked">>, Context1)}
```

src/modules/cb_user_auth.erl
```diff
@@ -796,9 +796,10 @@ error_no_account_name(Context, AccountName) ->
           kz_either:either(cb_context:context(), cb_context:context()).
 consume_tokens(Context) ->
     TokenCost = cb_modules_util:token_cost(Context, ?USER_AUTH_TOKENS),
-    case cb_modules_util:consume_tokens_until(Context, TokenCost) of
```
icehess, line 799: First call regular consume_token_until (which will do it job and check bucket for `client_ip/no_account_id`.
The call if true call consume_auth_tokens_until.

as i showed before something like this:

```erlang
    case cb_modules_util:consume_tokens_until(Context, TokenCost, BucketName) of
        {'true', Context1} ->
            AuthBucketNames = cb_modules_util:auth_bucket_names(Context1, []),
            case cb_modules_util:consume_auth_tokens_until(Context1, AuthBucketNames, Cost) of
                'true' -> ....
                'false' ->
                    {'error', cb_context:add_system_error(401, 'invalid_credentials', <<"too many failed login attempt, your account is locked">>, Context1)}
```

icehess (changes requested):

src/crossbar.hrl
```diff
@@ -9,6 +9,7 @@
 
 -define(APP, 'crossbar').
 -define(APP_NAME, <<"crossbar">>).
+-define(CROSSBAR_AUTH, <<"crossbar_auth">>).
```
icehess: The name of macro is too generic and confusing. it could mean anything. rename it to something like `CROSSBAR_AUTH_BUCKET`

src/cb_modules_util.erl
```diff
@@ -162,6 +163,29 @@ parse_media_type(MediaType) ->
             {'error', 'badarg'}
     end.
 
+-spec consume_auth_tokens_until(cb_context:context(), non_neg_integer()) -> boolean().
+consume_auth_tokens_until(Context, TokenCost) ->
+  BucketNames = auth_bucket_names(Context),
+  maybe_consume_auth_tokens_until(TokenCost, BucketNames).
+
+maybe_consume_auth_tokens_until(_TokenCost, []) -> 'true';
```
icehess: Export this function, we need it to allow modules call this with their own list of bucket names.

src/modules/cb_basic_auth.erl
```diff
@@ -57,9 +57,13 @@ authenticate(Context) ->
           'false' |
           {'true' | 'stop', cb_context:context()}.
 authenticate(Context, 'basic') ->
-    case cb_modules_util:consume_tokens(Context, cb_modules_util:token_cost(Context)) of
+    TokenCost = cb_modules_util:token_cost(Context),
+    case cb_modules_util:consume_tokens(Context, TokenCost) of
         {'true', Context1} ->
-            check_basic_token(Context1, cb_context:auth_token(Context1));
+            case cb_modules_util:consume_auth_tokens_until(Context1, TokenCost) of
```
icehess: This doesn't work here. Basic auth doesn't have any payload and `cb_modules_util:consume_auth_tokens_until/1` can't get account id.

But we do have account id. If you look further down we extract basic auth token and decode it which gives us a `[AccountId, Credentials]`.

You can refactor a bit here and do this decoding to extract any account id, if you found it then call the newly exported `cb_modules_util:consume_auth_tokens_until/3` (see the comment above) with the account id as bucket names.

icehess (changes requested):

src/modules/cb_security.erl
```diff
@@ -72,11 +73,21 @@ init() ->
 authorize(Context) ->
     authorize_list_available_module(Context, cb_context:req_nouns(Context), cb_context:req_verb(Context)).
 
--spec authorize(cb_context:context(), path_token()) -> 'true'.
-authorize(_Context, _) -> 'true'.
+-spec authorize(cb_context:context(), path_token()) -> 'true' | {'stop', cb_context:context()}.
```
icehess:
```suggestion
-spec authorize(cb_context:context(), path_token()) -> boolean() | {'stop', cb_context:context()}.
```

icehess (changes requested):

icehess (approved):

icehess:
Required PRs:
- https://github.com/2600hz/kazoo-crossbar/pull/467
- https://github.com/2600hz/kazoo-core/pull/663
- https://github.com/2600hz/kazoo-properly/pull/150
- https://github.com/2600hz/kazoo-desktop/pull/25

src/crossbar_auth.erl
```diff
@@ -508,3 +514,88 @@ refresh_token(Context) ->
...
+        Value <- [kz_json:get_ne_binary_value(Key, ReqData)],
+        kz_term:is_not_empty(Value)
+    ].
+
+-spec is_account_locked(cb_context:context(), [kz_term:ne_binaries()]) ->
+          {boolean(), cb_context:context()}.
+is_account_locked(Context, []) ->
+    {'false', Context};
+is_account_locked(Context, BucketNames) ->
+    case kapps_config:is_true(?AUTH_CONFIG_CAT, <<"lock_account_on_failed_attempts">>) of
```
icehess: `kapps_config:is_true/2` could return `undefined`, either handle it in the case clause or use arity 3 and set a default value to `false`.

icehess (changes requested):

priv/couchdb/schemas/system_config.crossbar.auth.json
```diff
@@ -31,7 +31,7 @@
                     "scopes": [
                         "crossbar"
                     ],
-                    "token_auth_expiry_s": "infinity"
+                    "token_auth_expiry_s": 3600
```
jamesaimonetti, line 34: Your properly tests couldnt pass because the API to update these configs failed due to this setting being a string instead of integer() and `cb_conference_auth` missing entirely (see below change).

dhuynh-Kazoo: hi @jamesaimonetti, following this I have a question: Does the change make the public conference available within one hour? As I saw this PR - [KCRO-104: Avoid conferences public links to expire](https://github.com/2600hz/kazoo-crossbar/pull/199/files), it changes `token_auth_expiry_s` to `infinity` for that purpose

jamesaimonetti: Good question, I don't know. I do know the schema expects an integer() and not "infinity". Clusters can configure this setting to a year or something, to be compliant with the schema, I guess.

dhuynh-Kazoo:
Hi @jamesaimonetti , I have a concern with cb_conference_auth. I don't really touch the `token_auth_expiry_s` of the module but if I do not change the scheme, it gets validation failed.

dhuynh-Kazoo:
hi @jamesaimonetti, could you review it again? Changes were updated

dhuynh-Kazoo:
Backport:

[[5.4] KZOO-301: add ability to stop token bucket](https://github.com/2600hz/kazoo-core/pull/995)
[[5.4]KZOO-301: Add a period of user lockout for failed login attempts](https://github.com/2600hz/kazoo-crossbar/pull/624)

PR #468, opened 2023-04-18
KFAX-23 : Expose parameters which required to determine file upload file size

wdhammika:
Previously, Support did not know what the monster UI should be updated to,  Currently, we have 3 different answers, comm.land states 15mb, the documentation states 3mb, and eFax in Kazoo report 20 MB. We are trying to track down the correct size

Two configuration parameters have been identified which determine file size.
```
crossbar => max_upload_size
fax => smtp_max_msg_size
```
Therefore `v2/accounts/{ACCOUNT_ID}/configs` has been configured to return parameters that UI needed to determine the file size and others. So UI can access those values through API.


src/modules/cb_configs.erl
```diff
@@ -43,16 +43,24 @@ init() ->
     _ = crossbar_bindings:bind(<<"*.execute.patch.configs">>, ?MODULE, patch),
     _ = crossbar_bindings:bind(<<"*.execute.delete.configs">>, ?MODULE, delete).
 
+-spec allowed_methods() -> http_methods().
+allowed_methods() ->
+    [?HTTP_GET, ?HTTP_POST, ?HTTP_PUT, ?HTTP_PATCH, ?HTTP_DELETE].
```
jamesaimonetti: I don't think this API should allow anything but GET, right?

jamesaimonetti (changes requested):
Want to see properly test too, to verify response data

src/modules/cb_configs.erl
wdhammika: Yes, GET is the main intention of this, but I have enabled other options to save/update parameters.

jamesaimonetti: These configs are global in nature and no account should be able to modify them. Please remove all but GET

wdhammika: Sure! I will remove them.

priv/api/swagger.json
```diff
@@ -397,6 +397,22 @@
             },
             "type": "object"
         },
+        "account_config.default": {
+            "description": "Schema for default account_config",
+            "properties": {
+                "max_upload_size": {
+                    "default": 20,
+                    "description": "maximum upload file size",
```
harenson, line 405: What measurement unit is being used? b, Kib, Mib, Gib, etc? Please, include it within the description. Something like: `"maximum (in {MEASURE}) upload file size"`.

wdhammika: Fixed.

doc/ref/configs.md
```diff
@@ -6,6 +6,16 @@
 
 
 
+## Fetch
```
jamesaimonetti, line 9: Bring this doc change down a dir to the `doc/configs.md` and add a sample response JSON payload

wdhammika: This is completed.

src/modules/cb_configs.erl
wdhammika: Added Properly test cases.
https://github.com/2600hz/kazoo-properly/pull/147

PR #470, opened 2023-04-20
KCRO-182: Cleanup account move

jamesaimonetti:
Previously, the post/3 would check cb_context for success (which used to carry over from validation but now resets between steps to error), and not finding success, would fail the request with a 500.

Now, ignore the status of cb_context and attempt the move.

Additionally, verify that the To account exists before updating the target account's tree.

Additional helpful logging when errors occur are added too.

Note: testing for this changeset centered around a superduper admin performing the API request. Account tree changes are not tested in Properly (yet).


jamesaimonetti:
Test in https://github.com/2600hz/kazoo-properly/pull/146

nkarunarathne:
These changes looks good. Properly test ran without any issue.

PR #474, opened 2023-04-26
KZOO-375: process optional timestamp against temporal rule

jamesaimonetti:
Return whether the rule matches for a given Gregorian timestamp. Helpful for checking that the rule matches (or not) expected timestamps without having to place calls to do so.


jamesaimonetti:
Check test here: https://github.com/2600hz/kazoo-properly/pull/148

harenson (approved):
:+1:

PR #478, opened 2023-05-18
KCRO-184: add owner_id to match lists and temporal rules schemas

jamesaimonetti:
Allow associating match lists and temporal rules with a user. Useful when setting up conditional call forwarding settings (among other reasons).


jamesaimonetti:
Requires changes from https://github.com/2600hz/kazoo-core/pull/671

PR #481, opened 2023-05-26
KCRO-183: multiple functions to handle all possible scenarios for direction

kusalfernando:
The reason for this defect was the 'children' were fetched in descending order which is the default and it should be overridden as ascending to fix this issue.
There was a bug in the API that ascending  defined in view options or query strings was not handle properly. It is fixed under this pull request.


src/crossbar_view.erl
```diff
@@ -742,7 +742,7 @@ direction(Context, Options) ->
     case props:get_value('direction', Options) =:= 'descending'
         orelse props:is_true('descending', Options, 'false')
         orelse kz_json:is_true(<<"descending">>, cb_context:query_string(Context))
-        orelse kz_json:is_false(<<"ascending">>, cb_context:query_string(Context), 'true')
+        orelse kz_json:is_false(<<"ascending">>, cb_context:query_string(Context), 'false')
```
icehess: Why this change? this change all APIs to default to ascending and breaks all other APIs, just set descending: true or direction: descnding in view options in cb_accounts.

icehess (changes requested):

src/crossbar_view.erl
icehess: Also imo UI should change their API calls and add ascending=true to qs. UI must follow the APIs rules not APIs follow what UI dictates.

kusalfernando: I have already tried with `{'direction', 'ascending'}` in view options and still comes to `orelse kz_json:is_false(<<"ascending">>, cb_context:query_string(Context), 'true')` and return default value of `descending`.

kusalfernando: According to my understanding, I guess ascending view option is not handled under case clause.

icehess: Yes it is not handling this properly. Can you fix the real issue? handle direction option first, if undefined handle descending boolean, if undefined handle QS options in that order, then if everything is undefined default to `descending`.


As for the ticket ask UI team to fix their API call to `/children` when listing accounts to include `ascending=true`. This should not be handled it backend. backend uniformly always returns result in descending order (unless there is a technical reason if that is is not possible) and offers ways to user to change the order if they want.

kusalfernando: got it. I will address the real problem and ask for the review.

kusalfernando: I made the changes and would like to get a review. Basic plan of the change is:
- handle {'direction', 'ascending'} or  {'ascending', 'true'} in the view options
- handle {<<"ascending">>, 'true'} in query string
- otherwise return default 'descending' as direction

icehess: Break this to multiple functions maybe, call each one at the time if they return a descending or ascending stop and return that if not continue to next function.

One function to handle `direction` in view option CHECK both descending if exists then ascending if exists, another fun to handle existence of `descending` in view option, another func to check for both descending if exists then ascending if exists in query string. then if no fun return anything, default to descending

kusalfernando: refactored direction/2 into multiple functions to handle all possible scenarios. Assumed 'ascending' 'false' is equal to 'descending' and 'descending' 'false' is equal to 'ascending'.

icehess (approved):

PR #489, opened 2023-06-14
Allow additional properties for system_config.crossbar.auth.json

dhuynh-Kazoo:
- Allow other properties do not match the default definition. 

- Dot adding cb_conference_auth to `properties` because `token_auth_expiry_s: infinity`


dhuynh-Kazoo:
Hi @icehess and @jamesaimonetti, I create the PR for the `additionalProperties` variable, please review it.

PR #494, opened 2023-06-21
KZOO-397: update schema for expected format of classifiers

jamesaimonetti:
Include required classifier properties


jamesaimonetti:
Requires https://github.com/2600hz/kazoo-core/pull/678

PR #495, opened 2023-06-22
KPUS-6 Add kazoo device id and message type into push payload.

wdhammika:
Previously, the push message payload included no push message type and Kazoo device id. Therefore this development is related to adding those parameters to push payload.

Preconditions: The device should have the push capability and it should be enabled.
An application should have registered in Pusher in order to capture push requests on the device.

Push messages can be cast as voicemail/missed_call/user_push and incoming call

Relevant PRs
https://github.com/2600hz/kazoo-pusher/pull/35


nkarunarathne:
This looks good. Syntax and code formatting are in order.

src/modules/cb_pushes.erl
```diff
@@ -271,3 +272,7 @@ maybe_ucfirst(Word, Options) ->
         Explicit -> Explicit
     end.
 %% }
+
+-spec add_push_type(kz_json:object()) -> kz_json:object().
+add_push_type(DataJobj) ->
```
Navoda626: It's better if we can use 'O' in DataJObj

Navoda626:
This looks good

PR #499, opened 2023-07-11
rebar3 as alternative experimental build tool

icehess:
This introduce rebar3 as an alternative build system, only for use locally in your dev
env for now.

Current Makefile system is robust and gets the job done. But it increasingly getting
difficult to run this in different env and OS, and it is ridiculously SLOW. Average time
is something around 7mins. If you run in VM it may be more close to 10min. If you add
macOS in equation, it is infinity (something around 25-35m min if the compile does not failed!).

With Rebar3 the build time is something under or around 2min, for everything, including deps.

compile, xref, eunit, fetching deps and release work as expected. Dialyzer reports all warning and
we can't ignore false-posstive currently.

You still need to fetch core and apps by `make fetch-core fetch-apps`.

This tries to not depend on calling Makefile targets as much as possible. This reduce the
problems on different platforms especially windows and macos. Some scripts was added to
replace some tasks needed to setup files for compilation.


icehess:
Required PRs:

- https://github.com/2600hz/kazoo5/pull/117
- https://github.com/2600hz/kazoo-crossbar/pull/499
- https://github.com/2600hz/kazoo-properly/pull/166
- https://github.com/2600hz/kazoo-pusher/pull/32
- https://github.com/2600hz/kazoo-registrar/pull/34
- https://github.com/2600hz/kazoo-tasks/pull/36
- https://github.com/2600hz/kazoo-teletype/pull/62
- https://github.com/2600hz/kazoo-core/pull/684

PR #500, opened 2023-07-11
[master] kcro-188-ListsApiEnhancements

kdamas93:
The purpose of this PR is to improve/adapt the current functionality of **cb_list** to be able to save contact lists depending on the needs, with these changes it will be possible to associate **owner_id** to the lists so that they can later be filtered by user.
With this **owner_id** permission, validation logic will be added, about who can access the lists.
Multiple entries can be associated to a **list** type document, each entry will be a contact list, to achieve this it is necessary to update the current schema (**list_entry**), the documentation and its respective accessors.
It is also necessary to perform a refactor on the current code because there are functions/calls that do not follow our current programming patterns.
According to the initial meeting, it is not necessary to contemplate or carry out a treatment for the photos or vcard methods


priv/couchdb/schemas/list_entries.json
```diff
@@ -3,57 +3,90 @@
...
-        },
-        "capture_group_length": {
-            "description": "Length of the numbers in the list entries for which it used to select desire Caller ID",
-            "type": "integer"
+        "contacts": {
+            "default": [],
+            "items": {
+                "description": "Contacts to save",
+                "properties": {
+                    "contact": {
```
jamesaimonetti, line 11: needs a `description` field

priv/couchdb/schemas/list_entries.json
```diff
@@ -3,57 +3,90 @@
...
-        },
-        "profile": {
-            "$ref": "profile",
+        "organization": {
             "default": {},
-            "description": "Profile data",
+            "description": "The organization this entity belongs to",
+            "properties": {
+                "name": {
+                    "default": "",
```
jamesaimonetti: no need for an empty default here, having `name` in the required section will ensure it is set to something

src/modules/cb_lists.erl
```diff
@@ -37,22 +39,19 @@
...
-                        ,{<<"*.resource_exists.lists">>, 'resource_exists'}
-                        ,{<<"*.content_types_provided.lists">>, 'content_types_provided'}
-                        ,{<<"*.validate.lists">>, 'validate'}
-                        ,{<<"*.execute.put.lists">>, 'save'}
-                        ,{<<"*.execute.post.lists">>, 'save'}
-                        ,{<<"*.execute.patch.lists">>, 'save'}
-                        ,{<<"*.execute.delete.lists">>, 'delete'}
-                        ]
-    ].
+    _ = crossbar_bindings:bind(<<"*.allowed_methods.lists">>, ?MODULE, 'allowed_methods'),
```
jamesaimonetti: Look at how cb_devices:init/0 does this, I would prefer the code is refactored in that direction.

src/modules/cb_lists.erl
```diff
@@ -210,6 +316,54 @@ validate_doc(Id, Type, Context) ->
     OnSuccess = fun(C) -> on_successfull_validation(Id, Type, C) end,
     cb_context:validate_request_data(type_schema_name(Type), Context, OnSuccess).
 
+-spec maybe_admin(cb_context:context()) -> boolean().
+maybe_admin(Context) ->
+    case cb_context:is_account_admin(Context)
```
jamesaimonetti: No need for the case expression:

```erlang
maybe_admin(Context) ->
    cb_context:is_account_admin(Context)
        orelse cb_context:is_superduper_admin(Context).
```

src/modules/cb_lists.erl
```diff
@@ -210,6 +316,54 @@ validate_doc(Id, Type, Context) ->
...
+        'false' -> 'false'
+    end.
+
+-spec remove_lists_with_owner_id(list(), list())-> list().
+remove_lists_with_owner_id([], Acc) ->
+    Acc;
+remove_lists_with_owner_id([List| Lists], Acc) ->
+    case kzd_lists:owner_id(List) of
+        'undefined' ->
+            remove_lists_with_owner_id(Lists, Acc ++ [List]);
```
jamesaimonetti: If you find yourself reaching for the append `++` operator, consider if there's a way to build the list differently. Prepending an item to a list, in Erlang, is super fast `[Item | List]` boom, done.

Appended to the end of a list means traversing `List` to add `Item` to the end, so the longer `List` is, the longer `++` takes. So this code is the absolute slowest way to build the list.

Instead, think about what this function wants to do - remove items from the list where an owner ID is set. Remove = filter. Two ways I would write that, then:

```
lists:filter(fun(List) -> kzd_lists:owner_id(List) =:= 'undefined' end, Lists)
```
This says, for each `List` in `Lists`, if the `owner_id` of `List` is undefined, return true (keep `List`); if `owner_id` returns something else, retrurn `false` (remove `List). Order of `Lists` is preserved for elements that remain.

Another option:
```
[List || List <- Lists, 'undefined' =:= kzd_lists:owner_id(List)].
```
Same thing, except as a list comprehension. Order is preserved.

src/modules/cb_lists.erl
```diff
@@ -210,6 +316,54 @@ validate_doc(Id, Type, Context) ->
...
+        _OwnerId ->
+            remove_lists_with_owner_id(Lists, Acc)
+    end.
+
+-spec filter_lists_by_owner_id(list(), list(), kz_term:api_ne_binary()) ->list().
+filter_lists_by_owner_id([], Acc, _UserId) ->
+    Acc;
+filter_lists_by_owner_id([List| Lists], Acc, UserId) ->
+    OwnerId = kzd_lists:owner_id(List),
```
jamesaimonetti: Seeing the 'remove' choices above, please refactor this to not use `++` too

src/modules/cb_lists.erl
```diff
@@ -210,6 +316,54 @@ validate_doc(Id, Type, Context) ->
...
+    case OwnerId =:= UserId of
+        'true' ->
+            filter_lists_by_owner_id(Lists, Acc ++ [List], UserId);
+        'false' ->
+            filter_lists_by_owner_id(Lists, Acc, UserId)
+    end.
+
+-spec get_url_user_id(cb_context:context()) -> kz_term:api_ne_binary().
+get_url_user_id(Context) ->
+    case props:get_value(<<"users">>, cb_context:req_nouns(Context)) of
```
jamesaimonetti: Check and see if `cb_context:user_id(Context)` returns the expected UserId

src/modules/cb_lists.erl
```diff
@@ -210,6 +316,54 @@ validate_doc(Id, Type, Context) ->
...
+        [UserId|_] -> UserId;
+        _ -> 'undefined'
+    end.
+
+-spec auth_token_user_matches(cb_context:context(), kz_term:api_ne_binary())-> boolean().
+auth_token_user_matches(Context, UserId)->
+    io:format("~p",[cb_context:auth_user_id(Context)]),
```
jamesaimonetti: Why `io:format`? Debug line that needs removing?

src/modules/cb_lists.erl
```diff
@@ -210,6 +316,54 @@ validate_doc(Id, Type, Context) ->
...
+    AuthUserId = cb_context:auth_user_id(Context),
+    case AuthUserId =:= UserId of
```
jamesaimonetti: No need for the case expression

src/modules/cb_lists.erl
```diff
@@ -233,34 +387,37 @@ delete(Context, _ListId, ?ENTRIES) ->
...
+-spec patch(cb_context:context(), path_token(), path_token(), path_token()) -> cb_context:context().
+patch(Context, _ListId, ?ENTRIES, _EntryId) ->
+    crossbar_doc:save(Context).
 
 -spec type_schema_name(kz_term:ne_binary()) -> kz_term:ne_binary().
 type_schema_name(?TYPE_LIST) -> <<"lists">>;
 type_schema_name(?TYPE_LIST_ENTRY) -> <<"list_entries">>.
 
 -spec on_successfull_validation(kz_term:api_binary(), kz_term:ne_binary(), cb_context:context()) -> cb_context:context().
 on_successfull_validation('undefined', Type, Context) ->
```
jamesaimonetti: One L in successful (i know its not your typo but we may as well fix it :)

jamesaimonetti (changes requested):

priv/couchdb/schemas/list_entries.json
kdamas93: `description `field added!

priv/couchdb/schemas/list_entries.json
kdamas93: change applied!

src/modules/cb_lists.erl
kdamas93: cb_lists:init/0 refactored based on suggested module

src/modules/cb_lists.erl
kdamas93: change applied!

src/modules/cb_lists.erl
kdamas93: Thank you very much for this, I found it very useful, it has made it much easier for me, thanks for the explanation, change applied!

src/modules/cb_lists.erl
kdamas93: change applied!

src/modules/cb_lists.erl
kdamas93: I proceeded to use `cb_context:user_id(Context)` instead of the implemented function, which is unnecessary because the one mentioned above already exists.

src/modules/cb_lists.erl
kdamas93: removed line

src/modules/cb_lists.erl
kdamas93: change applied!

src/modules/cb_lists.erl
kdamas93: change applied!

src/modules/cb_lists.erl
```diff
@@ -145,48 +147,155 @@ validate(Context, ListId, ?ENTRIES, EntryId, ?VCARD) ->
...
+              ,'include_docs'
+              ],
+    RespContext = crossbar_view:load(Context, ?CB_LIST, Options),
+    RespDoc = cb_context:resp_data(RespContext),
+    case cb_context:user_id(Context) of
+        'undefined' ->
+            NewDoc = [List || List <- RespDoc, 'undefined' =:= kzd_lists:owner_id(List)],
+            cb_context:set_resp_data(RespContext, NewDoc);
+        UserId ->
+            NewDoc = [List || List <- RespDoc, UserId =:= kzd_lists:owner_id(List)],
```
icehess: Doing filtration in this way will/may breaks the pagination that `crossbar_view` tries so hardly to maintain.

I suggest either write a proper `mapper` function which checks this ownership or even create a special view `list_by_ownerid`.

You can even use a generic view like `crossbar_listings/by_ownerid_type` and use view option `{key, [UserId, <<"lists">>]}`

src/modules/cb_lists.erl
```diff
@@ -145,48 +147,155 @@ validate(Context, ListId, ?ENTRIES, EntryId, ?VCARD) ->
...
 validate_req(?HTTP_GET, Context, []) ->
-    crossbar_view:load(Context, ?CB_LIST, [{'mapper', crossbar_view:get_value_fun()}]);
+    Options = [{'mapper', crossbar_view:get_doc_fun()}
+              ,'include_docs'
+              ],
+    RespContext = crossbar_view:load(Context, ?CB_LIST, Options),
+    RespDoc = cb_context:resp_data(RespContext),
+    case cb_context:user_id(Context) of
+        'undefined' ->
+            NewDoc = [List || List <- RespDoc, 'undefined' =:= kzd_lists:owner_id(List)],
```
icehess: This potentially breaks the pagination, please consider to either use a proper `mapper` to filter out unwanted result or use special view to `list_unowned`.

Also considering the comment in the next `case` clause, if you are using different views to list based on existence of user id, and then please move these to their own function and call `crossbar_view:load_view` with their oen view and options.

src/modules/cb_lists.erl
```diff
@@ -145,48 +147,155 @@ validate(Context, ListId, ?ENTRIES, EntryId, ?VCARD) ->
...
+            cb_context:set_resp_data(RespContext, NewDoc);
+        UserId ->
+            NewDoc = [List || List <- RespDoc, UserId =:= kzd_lists:owner_id(List)],
+            cb_context:set_resp_data(RespContext, NewDoc)
+    end;
 validate_req(?HTTP_PUT, Context, []) ->
-    validate_doc('undefined', ?TYPE_LIST, Context);
+    case cb_context:user_id(Context) of
```
icehess, line 239: Instead of checking authorization in validate function how about use `authorize` callback and match on method and nouns/user id?

icehess (changes requested):

src/modules/cb_lists.erl
kdamas93: proceeded to create a custom mapper for unowned lists and for owned lists a crossbar view is used

src/modules/cb_lists.erl
kdamas93: proceeded to create a custom mapper for unowned lists and for owned lists a crossbar view is used, In the same way, it was separated into different functions since each one uses a different view.

src/modules/cb_lists.erl
kdamas93: the logic of who can access the resources was implemented in the authorize methods

PR #504, opened 2023-07-18
KCRO-171: Implementation of kazoo teammate application

Navoda626:
Adding configs related to teammate application.


icehess:
If you set `{schemas_to_priv,true}` in your `.app.src` env key you won’t need this pr anymore. ast will put the schemas inside the app's repo.

Navoda626:
@icehess Yeah got it. I updated the teammate with `{schemas_to_priv,true}`, So removing the PR

PR #506, opened 2023-07-25
kcro-148-RefactorAppExhange

kdamas93:
At the moment App Exchange does not show the applications that do not have entitlements for the admins, this is necessary so that they can create the entitlement and thus other users can see it in their account depending on the assigned entitlement.
This is a small fix in relation to these changes:
https://github.com/2600hz/kazoo-crossbar/pull/374


icehess:
Properly test: https://github.com/2600hz/kazoo-properly/pull/173

icehess (approved):

PR #508, opened 2023-07-27
KCRO-189: allow desktop endpoint to have an optional authentication to access comm.land artifacts and use whitelabel document if it exists

Navoda626:
Adding new parameters in whitelabel document `commland.downloads.mac`, `commland.downloads.windows`, `commland.downloads.deb` and `commland.downloads.rpm` to facilitate Account specific URLs for downloading commland artifacts. 

Adding `allow_public_downloads` system configuration to bypass the authentication and authorization for requests for crossbar API.


nkarunarathne:
Code for mating and syntax are in order.

kdamas93:
I think it would be a good practice to unify commits using git squash

PR #513, opened 2023-08-11
[master] SUPP-322: Numbers not found when trying to delete them from Numbers app

dhuynh-Kazoo:


src/modules/cb_phone_numbers.erl
```diff
@@ -1020,3 +1028,19 @@ default_knm_options(Context) ->
...
+
+%%------------------------------------------------------------------------------
+%% @doc
+%% Check if the number id in account db exist in prefix number db
+%% @end
+%%------------------------------------------------------------------------------
+-spec check_number_in_prefix_db(tuple()) -> boolean().
+check_number_in_prefix_db({Number, _Attrs}) ->
+    Db = knm_converters:to_db(Number),
+    Doc = kz_datamgr:open_cache_doc(Db, Number),
```
icehess, line 1040: Do not interact with numbers directly like this. kazoo_number is the only place that abstract all things related to numbers.

Looking at the ticket comments on the ticket, it seems the account db and number db got out of sync. Find the root cause if you can but asking questions and examine and find out why dbs ended up in this state.

THis happened a lot previously, so we have a sync mechanism that check number dbs and fixes this uses

icehess (changes requested):

src/modules/cb_phone_numbers.erl
dhuynh-Kazoo: thanks @icehess!! I will check the maintenance file.

PR #516, opened 2023-09-22
KZOO-410: UUID treatment for command selector

kdamas93:
Is necessary implement/automate the use of UUID for the command selector for its _id.
To do this, it is necessary to receive the uuid field in the payload through the console and in the URL from the API and this will be validated and stored in the _id and uuid fields documents.


Navoda626:
@kdamas93 Its better to change the PR heading and other than that changes looks good

PR #519, opened 2023-09-28
[5.3] KCRO-194: Fix double url encoding for chat

kdamas93:
Fix the double encoding that the endpoint produces and which was noticed by the version of Erlang used
FYI :
image


Navoda626:
Code and syntax seems good

PR #522, opened 2023-10-03
KCRO-195 : As a user I want to update date from the Monster UI entry for Marketplace connector settings

manushiperera:
Previously it was unable to set `is_aio_cluster` configuration through `PATCH apps_store/marketplace` API. Now this configuration can be set by `PATCH apps_store/marketplace: {"action":"enable", "settings":["is_aio_cluster",…]}`. `settings` is an optional parameter.


src/modules/cb_apps_store.erl
```diff
@@ -431,6 +433,22 @@ patch_market_action(Context, ?MARKET_ACTION_API_URL) ->
     _ = appex_client_config:set_api_url(ApiUrl),
     crossbar_doc:handle_datamgr_success(appex_client_config:get_public_configs(), Context).
 
+may_be_enable_aio_cluster('undefined') ->
+    'ok';
+may_be_enable_aio_cluster(Settings) ->
+    case lists:member(<<"is_aio_cluster">>, Settings) of
```
icehess: Treat `Settings` as a JObj with a payload like this:

```json
{
    "action": "....",
    "settings": {
        "is_aio_cluster": boolean,
        "name": string
    }
}
```

src/modules/cb_apps_store.erl
```diff
@@ -400,11 +400,13 @@ patch(Context, ?MARKETPLACE) ->
 
 patch_market_action(Context, ?MARKET_ACTION_ENABLE) ->
     _ = appex_client_config:set_is_enabled('true'),
+    may_be_enable_aio_cluster(cb_context:req_value(Context, <<"settings">>)),
     maybe_restart_kazoo_appex(),
     crossbar_doc:handle_datamgr_success(appex_client_config:get_public_configs(), Context);
 patch_market_action(Context, ?MARKET_ACTION_DISABLE) ->
     %% don't stop kazoo_appex to let it clean out apps when they stop
     _ = appex_client_config:set_is_enabled('false'),
+    may_be_disable_aio_cluster(cb_context:req_value(Context, <<"settings">>)),
```
icehess: Instead of disable/enable function have a function which parses settings (if defined) and set the name or is_aio_cluster value if they present in the jobj.

Also call this function for these actions: enable/link/api_url/update. So the UI can set any of these values in those steps if needed.

icehess (changes requested):
Some notes about the shape of `settings` object.

Can you write a simple doc for this in `apps_store.md`? List of api endpoints and a sample request payload/response would be very nice.

src/modules/cb_apps_store.erl
manushiperera: Done

src/modules/cb_apps_store.erl
manushiperera: Done

icehess:
@manushiperera I did some refactoring, to explicitly handle what setting is local and what needs to be update in appex server. There is another core pr which is required by this PR: https://github.com/2600hz/kazoo-core/pull/726

I didn't test this code, please take a time an test this. Please review the new code and fix any CI issues if there is any. If you can also please proof read the doc, I usually have a lot of typos :))

doc/apps_store.md
```diff
@@ -488,3 +488,223 @@ curl -v -X DELETE \
...
+
+If the sysadmin decided to unlink their cluster from Marketplace, they can use this API to
+do so.
+
+Please note this won't stop the currently running apps that are fetched off marketplace.
+Also, your purchases in your Marketplace Account won't be affected and you are going to
+billed for them.
+
+
+### Setting Crossbar API URl
```
icehess: On second thought let's remove this, and instead of update UI to use the new update action.

src/modules/cb_apps_store.erl
```diff
@@ -421,37 +422,64 @@ patch_market_action(Context, ?MARKET_ACTION_UNLINK) ->
     case appex_client_client:unlink_cluster() of
         {'ok', _RespData} ->
             _ = appex_client_config:set_is_enabled('false'),
-            crossbar_doc:handle_datamgr_success(appex_client_config:get_public_configs(), Context);
+            send_marketplace_configs(Context);
         {'error', Error} ->
             cb_context:add_system_error(500, <<"appex_passthrough_error">>, Error, Context)
     end;
 patch_market_action(Context, ?MARKET_ACTION_API_URL) ->
```
icehess: On second thought let's remove this, and instead of update UI to use the new update action.

src/modules/cb_apps_store.erl
```diff
@@ -663,32 +691,40 @@ validate_market_action(Context, ?MARKET_ACTION_ENABLE) ->
...
-        'undefined' ->
-            Message = kz_json:from_list([{<<"message">>, <<"access_code is required but it is missing">>}]),
-            cb_context:add_validation_error(<<"access_code">>, <<"required">>, Message, Context);
-        _ ->
-            set_ctx_success(Context)
-    end;
+    validate_market_settings(Context, <<"access_code">>);
 validate_market_action(Context, ?MARKET_ACTION_UNLINK) ->
     set_ctx_success(Context);
 validate_market_action(Context, ?MARKET_ACTION_API_URL) ->
```
icehess: remove this.

icehess (changes requested):

PR #524, opened 2023-10-10
[5.3]-KCRO-197: whitelabeled/blacklisted apps breaks the app selector list

kdamas93:
Adjusting instructions for managing applications that come from the service plan
When a service plan has a whitelabeled/blacklisted application generates a break since the structure returned is not what was expected, this exception is adapted with the added changes
error reported :
https://2600hz-commercial.atlassian.net/browse/CAS-110


icehess (approved):

icehess:
master pr: https://github.com/2600hz/kazoo-crossbar/pull/523

PR #529, opened 2023-10-31
[5.3] KCRO-188: Update contact lists (#500)

kdamas93:
The purpose of this PR is to improve/adapt the current functionality of cb_list to be able to save contact lists depending on the needs, with these changes it will be possible to associate owner_id to the lists so that they can later be filtered by user.

With this owner_id permission, validation logic will be added, about who can access the lists.

Multiple entries can be associated to a list type document, each entry will be a contact list, to achieve this it is necessary to update the current schema (list_entry), the documentation and its respective accessors.

It is also necessary to perform a refactor on the current code because there are functions/calls that do not follow our current programming patterns.

https://github.com/2600hz/kazoo-crossbar/pull/500


Navoda626:
@kdamas93 Its better to include the related PRs in the description and please remove the `(#500)` which maybe the master PR reference from the heading

These changes match with the master PR and the changes looks good

manushiperera:
Hello @kdamas93, 
Can you please share the related PRs for the same?

manushiperera:
Related PRs 
https://github.com/2600hz/kazoo-core/pull/738
https://github.com/2600hz/kazoo-properly/pull/186

icehess:
Please do not merge yet. There are concerns about impact of these changes, for all branches, including master. The PR changed `list_entry` docs fields like `number` and `pattern` but the code in other repos are not updated to handle these changes.

jamesaimonetti:
@icehess are we any closer to resolving the issues of this PR (or the master variant)?

icehess:
There was a meeting a while ago about this with some new suggestions and clarifications, I think this API is going to change to the new spec and we just try to use the new list structure with anything used in callflow modules.

PR #540, opened 2023-12-04
AXS-133:  add a cb module to handle ui wrapper app routes

icehess:
Currently UI asset wrapper app uses Crossbar Custom Route to add their routes
to the Crossbar. But this needs a restart of crossbar to take effect.

It is not useful to sysadmin as they start any UI app wrapper from
Marketplace and they expect its assets be available from a currently running
crossbar without any interruption to API availability.

This adds a new module to crossbar to handle `/ui/[...]` routes and
calls the apps wrapper module if it is exists.

The first part of path immediately after `ui` is the name of UI
wrapper app. The rest is the path to assets itself.

`crossbar_ui_wrapper_handler` will register itself to `/ui/[...]` in
cowboy router, and upon its calling, checks the app has a specific
module to handle this route is loaded then upgrades to that module for
handling the request.

For example:

`/ui/monsterui_callflows/[...]` will init `crossbar_ui_wrapper_handler`
then it tries to upgrade the handler to `monsterui_callflows_cb_router`
module if it exists. if not 404 will be returned.


icehess:
Kazoo PRs:
- https://github.com/2600hz/kazoo-crossbar/pull/540
- https://github.com/2600hz/kazoo-crossbar/pull/542

Related Changes:
- https://github.com/2600hz/bluepak-generators/commit/43aa5715b4fb3c5554c6213bdf021d98a9368db6
- https://github.com/2600hz/appex-bluepak/commit/704308416672532a56feb75a69b035c0894c1741

src/crossbar_ui_wrapper_handler.erl
```diff
@@ -0,0 +1,90 @@
...
+    case get_app_name(PathInfo) of
+        'undefined' ->
+            not_found(Req, HandlerOpts);
+        {Module, NewPathInfo} ->
+            {?MODULE, Req#{path_info => NewPathInfo}, [{'module', Module}]}
+    end.
+
+-spec upgrade(cowboy_req:req(), cowboy_middleware:env(), any(), any()) -> {'ok', cowboy_req:req(), cowboy_middleware:env()}.
+upgrade(Req, Env, _Handler, HandlerOpts) ->
+    Module = proplists:get_value('module', HandlerOpts),
```
jamesaimonetti: eww proplists

PR #541, opened 2023-12-05
[master] KCRO-200: Extension Reassignment Error Leads to Unintended Overwrite

filipevarjao:
Steps to Reproduce the error:

1. User A unassigns their current extension.
2. User A attempts to assign themselves extension 1001, which is already in use by User B.
3. Upon saving, User A receives an error message, indicating the assignment process failed.
4. However, after refreshing the system, it is observed that User A has been assigned extension 1001.
5. Concurrently, User B, the original holder of extension 1001, no longer has this extension assigned.

When assigning the user's extension we ignore the vmbox uniqueness validation, it fails at vmbox level but we store it to the user document anyway, adding the validation.


icehess:
Any reasons why this was not done in `kzd_users:validate/3` instead? or why patch/put did not check for this?

`kzd_users:validate/3` is the best place to check for uniqueness of presence id. It also covers for both PUT/PATCH/POST

src/modules/cb_users.erl
```diff
@@ -251,9 +251,14 @@ validate_users(Context, ?HTTP_PUT) ->
 validate_user(Context, UserId, ?HTTP_GET) ->
     load_user(UserId, Context);
 validate_user(Context, UserId, ?HTTP_POST) ->
-    UserJObj = kzd_users:maybe_migrate_user_addresses(cb_context:req_data(Context)),
-    Context1 = load_user(UserId, cb_context:set_req_data(Context, UserJObj)),
-    validate_request(UserId, Context1);
+    NewContext = validate_request(UserId, Context),
```
jamesaimonetti, line 254: This PR seems unnecessary now, no? Why validate the request data, then modify the request data, then validate the request data again?

jamesaimonetti:
@filipevarjao is this PR needed now?

filipevarjao:
No more needed, we moved all verification to the core.

PR #543, opened 2023-12-19
AXS-134: filter out ui apps started from marketplace if its app is not running

icehess:
When Apps Store API or the new Applications API is returning the list of allowed or available apps, they should now check if the app is from Marketplace, and if yes they have to check whether its app is started or not.

This will make sure these APIs only returning the apps that are accessible and running in the system.


icehess:
PRs:
- master: https://github.com/2600hz/kazoo-crossbar/pull/543
- 5.3: https://github.com/2600hz/kazoo-crossbar/pull/544

Related changes:
- https://github.com/2600hz/bluepak-generators/commit/758f7dbddb535e39d8a52b880a782dd151348160

src/cb_apps_util.erl
```diff
@@ -849,3 +851,88 @@ app_whitelabel_doc_id_to_app_id(<<"app_whitelabel-", AppId/binary>>) ->
...
+maybe_remove_kappex_apps(_AppType, AccountId, UserId, AppsMap) ->
+    maybe_remove_kappex_apps(AccountId, UserId, AppsMap).
+
+%%------------------------------------------------------------------------------
+%% @doc Get a list of all running app across the whole cluster.
+%% @end
+%%------------------------------------------------------------------------------
+-spec get_kapps() -> kz_term:ne_binaries().
+get_kapps() ->
+    lists:usort(
```
jamesaimonetti: IMO this is too many nested actions; hard to reason about what's happening and in what order.

Simple refactoring made it more readable for me:
```erlang
    ZoneProperties = [kz_json:values(ZoneNode) || ZoneNode <- kz_json:values(kz_nodes:status_to_json())],
    NodeProperties = [kz_json:values(NodeType) || NodeType <- lists:append(ZoneProperties)],
    RunningKapps = [kz_json:get_list_value(<<"kapps">>, Node, [])
                    || Node <- lists:append(NodeProperties)
                   ],
    lists:usort(lists:append(RunningKapps)).
```

Its mostly capturing intermediate values into named variables to help the reader see what data is being operated on.

src/cb_apps_util.erl
```diff
@@ -849,3 +851,88 @@ app_whitelabel_doc_id_to_app_id(<<"app_whitelabel-", AppId/binary>>) ->
...
+%% will not be listed in available apps and makes MUI or Commland App loaders
+%% only show current running apps.
+%% @end
+%%------------------------------------------------------------------------------
+-spec maybe_remove_kappex_apps(kz_term:ne_binary(), kz_term:api_ne_binary(), apps_map()) -> apps_map().
+maybe_remove_kappex_apps(_AccountId, _UserId, AppsMap) ->
+    Kapps = get_kapps(),
+    maps:fold(fun (AppId, AppJObj, AppsMapAcc) ->
+                      IsFromKappex = kz_json:is_true(<<"from_kappex">>, AppJObj),
+                      case maybe_remove_kappex_app(Kapps, AppJObj, IsFromKappex) of
```
jamesaimonetti: Perhaps the function should be named `should_remove_kappex_app`? `maybe_*` functions typically evaluate a boolean condition and do the thing if the condition is true, while `should_*` generally means evaluate a condition and return the boolean().

src/cb_apps_util.erl
icehess: no fun nerding? :))

PR #544, opened 2023-12-19
[5.3] AXS-134: filter out ui apps started from marketplace if its app is not running

icehess:
When Apps Store API or the new Applications API is returning the list of allowed or available apps, they should now check if the app is from Marketplace, and if yes they have to check whether its app is started or not.

This will make sure these APIs only returning the apps that are accessible and running in the system.


icehess:
PRs:
- master: https://github.com/2600hz/kazoo-crossbar/pull/543
- 5.3: https://github.com/2600hz/kazoo-crossbar/pull/544

Related changes:
- https://github.com/2600hz/bluepak-generators/commit/758f7dbddb535e39d8a52b880a782dd151348160

PR #549, opened 2024-02-06
KCRO-204: As an admin I want the ability to create a JWT that only grants read access

manushiperera:
Previously, it was unable to grant read only access for specific users. Now, if `pvt_scopes` field is defined in `user` doc with `fetch-exclusive`, creates the JWT token that contains a claim for read only access. When JWT is authorized, check for that claim and if it presents then the request is denied if the HTTP method is not a `GET`.


manushiperera:
Related PRs
https://github.com/2600hz/kazoo-core/pull/778
https://github.com/2600hz/kazoo-properly/pull/196

src/modules/cb_token_auth.erl
```diff
@@ -301,3 +301,32 @@ maybe_add_is_admins(<<AuthAccountId/binary>>, 'undefined') ->
...
+        Scopes ->
+            case maybe_validate_scope(Context, kz_auth_scope:to_list(Scopes)) of
+                'true' ->
+                    check_as(Context, JObj);
+                {'false', {Code, Error, Msg}} ->
+                    {'stop', cb_context:add_system_error(Code, Error, Msg, Context)}
+            end
+    end.
+
+-spec maybe_validate_scope(cb_context:context(), kz_term:api_binaries()) ->
```
icehess:
```suggestion
-spec maybe_validate_scope(cb_context:context(), [kz_term:ne_binary() | kz_term:ne_binaries()]) ->
```

src/modules/cb_user_auth.erl
```diff
@@ -359,13 +359,13 @@ maybe_include_claims(Context) ->
...
 
--spec maybe_fetch_scopes(cb_context:context()) -> kz_term:api_ne_binaries().
+-spec maybe_fetch_scopes(cb_context:context()) -> kz_term:api_binaries().
 maybe_fetch_scopes(Context) ->
     AccountId = kz_json:get_ne_binary_value(<<"account_id">>, cb_context:doc(Context)),
     OwnerId = kz_json:get_ne_binary_value(<<"owner_id">>, cb_context:doc(Context)),
     case kzd_users:fetch(AccountId, OwnerId) of
         {'ok', Doc} ->
-            kzd_users:scope_restrictions(Doc);
+            kzd_users:scope_restrictions(Doc, []) ++ kzd_users:pvt_scopes(Doc, []);
```
icehess, line 368: Any reason why not using `scope_restriction`?

Or any reason why using a private `pvt_scopes` field instead of publish one that can be easily added to user doc using API?

src/modules/cb_token_auth.erl
```diff
@@ -301,3 +301,32 @@ maybe_add_is_admins(<<AuthAccountId/binary>>, 'undefined') ->
...
+        'undefined' ->
+            check_as(Context, JObj);
+        Scopes ->
+            case maybe_validate_scope(Context, kz_auth_scope:to_list(Scopes)) of
+                'true' ->
+                    check_as(Context, JObj);
+                {'false', {Code, Error, Msg}} ->
+                    {'stop', cb_context:add_system_error(Code, Error, Msg, Context)}
+            end
+    end.
```
icehess: You can shorten the code like this:

```suggestion
    ExpandedScopes = kz_auth_scope:to_list(kz_auth_claims:scope(JObj, [])),
    case maybe_deny_access(Context, ExpandedScopes) of
        'false' ->
            check_as(Context, JObj);
        {'true', ContextError} ->
            %% set the error in context inside maybe_deny_access/2 and return:
            %% `true | {'true', ContextError}`
            {'stop', ContextError}
    end.
```

icehess (changes requested):

src/modules/cb_token_auth.erl
manushiperera: Done

src/modules/cb_user_auth.erl
manushiperera: `scope_restriction` is yet to be developed. Since this is an urgent requirement from Ooma, Karl advised to use `pvt_scopes`.

src/modules/cb_token_auth.erl
manushiperera: Done

lazedo:
looks like this can be closed ? see comments on https://github.com/2600hz/kazoo-core/pull/778 that should also be closed.

manushiperera:
Since we can use existing `scope_restriction` to cater this requirement, this PR is irrelevant.

First, it is required to define a scope restriction along with token restriction rule.
```
  "scopes": [
    "crossbar:read_only"
  ],
  "token_restrictions": {
    "_": [
      {
        "allowed_accounts": [
          "_"
        ],
        "rules": {
          "#": [
            "GET"
          ]
        }
      }
    ]
  }
```
Then we can use that doc id to define `scope_restrictions` in user doc.

```
{
        "last_name":"test1",
        "first_name": "test0",
        "password": "test1",
        "username": "test1",
        "scope_restrictions": [SCOPE_RESTRICTION_DOC_ID]
}
```

PR #551, opened 2024-02-21
[5.3] KTEL-21: As a reseller I want my emails to be signed with DKIM so that they are reliably delivered

manushiperera:
Previously, DKIM information is not used while sending emails. Now, private key file and DKIM selector information are sending through `whitelabel` API. A new API is defined to upload the private key file. If a file is already available, it is replaced. Otherwise the file is saved and put the file ID in `whitelabel` doc as a `pvt` field.


manushiperera:
Related PRs

https://github.com/2600hz/kazoo-teletype/pull/71
https://github.com/2600hz/kazoo-core/pull/784
https://github.com/2600hz/kazoo-properly/pull/197

icehess (approved):

icehess:
another pr fixing whitelabel.md

https://github.com/2600hz/kazoo-crossbar/pull/552

PR #554, opened 2024-02-29
KCRO-205: fix `cb_user_auth` always logging "found more than one user"

danielfinke:


src/modules/cb_user_auth.erl
```diff
@@ -533,30 +533,36 @@ maybe_account_is_enabled(Context, Account) ->
...
-load_sha1_results(Context, [JObj|_], _Account)->
-    lager:debug("found more that one user with SHA1 creds, using ~s", [kz_doc:id(JObj)]),
-    set_credentials_doc(Context, kz_json:get_value(<<"value">>, JObj));
 load_sha1_results(Context, [], Account)->
     Reason = io_lib:format("failed to find a user with SHA1 creds, request from IP address: ~s", [cb_context:client_ip(Context)]),
     lager:warning("~s", [Reason]),
     crossbar_auth:log_failed_auth(?MODULE, <<"credentials">>, kz_term:to_binary(Reason), Context, Account),
     cb_context:add_system_error('invalid_credentials', Context);
-load_sha1_results(Context, JObj, _Account)->
-    lager:debug("found SHA1 credentials belong to user ~s", [kz_doc:id(JObj)]),
```
lazedo, line 545: why don't you just change this to [JObj] and move it before line 534 ? no need for a `case` when pattern matching can handle ut

danielfinke: Fine by me 🙂

lazedo (approved):

icehess (approved):

PR #580, opened 2024-06-13
KCRO-206 : As a developer, I want an API for crud operations on a list of from addresses so that can DKIM to sign emails

manushiperera:
Previously, DKIM information was common for the account and unable to configure based on the available email addresses of the account. Now, private key file and DKIM selector information can be configured for each email address through new `whitelabel` API. A new crud operations are defined to configure, verify and update email addresses along with DKIM information.


manushiperera:
Related PRs
https://github.com/2600hz/kazoo-core/pull/867
https://github.com/2600hz/kazoo-teletype/pull/73
https://github.com/2600hz/kazoo-properly/pull/208

src/modules/cb_whitelabel.erl
```diff
@@ -621,17 +745,49 @@ validate_domain(Context, Domain, ?HTTP_GET) ->
...
+    lager:debug("created 4 digit pin code, sending email to '~s'", [Email]),
+    Notify = [{<<"Email">>, Email}
+             ,{<<"Verification-Code">>, Code}
+             ,{<<"Account-ID">>, kz_doc:account_id(Doc)}
+             ,{<<"Account-DB">>, kz_doc:account_db(Doc)}
+             | kz_api:default_headers(?APP_NAME, ?APP_VERSION)
+             ],
+    kapps_notify_publisher:cast(Notify, fun kapi_notifications:publish_email_verification/1),
+    Msg = <<"Email verification code was sent, Please check your email to learn how to verify your email address.">>,
+    crossbar_util:response(Msg, Context).
```
jamesaimonetti: The function returns a modified `cb_context` record but no calls to this function use the returned record. Seems like we can just return 'ok'?

manushiperera: Yeah that makes sense
Thanks James

PR #607, opened 2024-10-02
[5.4] KCRO-216: check for user's password expiration

jamesaimonetti:
Backport of #604

- include password expiration timestamp in user metadata
- set require_password_update to true if password is expired


jamesaimonetti:
Requires https://github.com/2600hz/kazoo-core/pull/975

PR #609, opened 2024-10-07
KCRO-218: Failure to Save Message Metadata Due to Incorrect Context Status Check in cb_vmboxes Module

wdhammika:


src/modules/cb_vmboxes.erl
```diff
@@ -601,19 +601,8 @@ check_mailbox_for_messages_array(Context, VMBoxMsgs) ->
...
-    case cb_context:resp_status(Context1) of
-        'success' ->
-            save_attachment(Context, cb_context:req_files(Context));
-        _ ->
-            Context1
-    end;
-maybe_save_attachment(Context, _Error) ->
-    lager:info("saving message metadata failed"),
-    Context.
+    save_attachment(Context1, cb_context:req_files(Context1)).
```
icehess, line 610: lgtm, just please check resp_status in `put/3` before calling `maybe_save_attachment/1`

icehess (changes requested):

src/modules/cb_vmboxes.erl
wdhammika: @icehess Could you please review the change?

icehess (approved):

lazedo:
@wdhammika @icehess I can be wrong but this pr changes nothing

icehess:
@lazedo there is also `put/5` calling `maybe_save_attachment/1` without any save operation and would fail because default context resp status is fatal.

lazedo:
@icehess so the `fix` is wrong, it should instead set rest_status on validation

icehess:
validation is fine, before execute request api_util sets resp status to fatal and all modules have to set it to success in their `put/post/delete/patch`. this module left this edge case and didn't set it.

lazedo:
isn't that what I'm saying ? the proper fix would have been to set resp_status. this all thing of putting request api_util sets resp status to fatal still hitting us, not a good change.

PR #611, opened 2024-10-08
[5.3] KCRO-218: Failure to Save Message Metadata Due to Incorrect Context Status Check in cb_vmboxes Module

wdhammika:
Backport of https://github.com/2600hz/kazoo-crossbar/pull/609


sanjeevie6353:
Reviewed and verified with the original PR.

PR #614, opened 2024-10-18
KCRO-225 Need to support Call Recording on demand

enoskov-at-ooma:
Add support to channels API to start and stop call recording. This is used by Ooma mobile and desktop clients.


enoskov-at-ooma:
@jamesaimonetti Can you please take a look?

src/modules/cb_channels.erl
```diff
@@ -611,3 +615,36 @@ get_account_id(Context) ->
             end;
         AccountId -> AccountId
     end.
+
+%%------------------------------------------------------------------------------
+%% @doc API action for starting or stopping call recording for a channel.
+%% For now, I do not check whether the call recording is enabled or not.
+%% @end
+%%------------------------------------------------------------------------------
+-spec maybe_start_record(cb_context:context(), kz_term:ne_binary()) -> cb_context:context().
```
dhuynh-Kazoo, line 624: You can try with `kapps_call:start_recording/2`

https://github.com/2600hz/kazoo-core/blob/717b9d105e9316e8a9c7a6c7a070bfb637f83cb4/kazoo_call/src/kapps_call.erl#L1813

lazedo: you can't call kapps_call without a call object

lazedo (approved):

PR #622, opened 2024-11-11
[5.4] KCRO-218 :handle vmbox media upload error

wdhammika:
Backport of https://github.com/2600hz/kazoo-crossbar/pull/609

(The original backport PR was reverted because of code merge freeze when the PR raised. Raising the PR again because the code freeze is not there anymore.)

Related PRs :
Master - https://github.com/2600hz/kazoo-crossbar/pull/609
5.3 - https://github.com/2600hz/kazoo-crossbar/pull/623


nkarunarathne (approved):
Looks good

PR #644, opened 2025-01-15
KPIV-53: basic Pivot serverless feature for processing static flow

icehess:
Callflow Pivot module can now accepts a serverless field and with a Pivot flow to be executed.

This removes the dependence of a web server and make using Pivot execute the static Pivot flow instead of making a HTTP request.

The new `serverless` field is an object with `flow_doc` and `flow_type` fields.

- `flow_doc`: can be string with a valid PivotML (a twiml extension) or a valid Callflow JSON
- `flow_type`: indicates the type of specified flow_doc. can be either `application/json` for callflow or `application/xml` or `text/xml` for PivotML.

Using `serverless` or `voice_url` is mutual exclusive, meaning they can't be used together. In serverless special take must be taken to use a full URL for any action url throughout the PivotML and relative URL will not work in serverless.

PivotML verbs that make a HTTP request for more instruction will not work in serverless fashion and still need a fully functional server.

Flow doc will be serialized to string if it is JSON and then will be send to Pivot app to be processed alongside the flow_type.

Pivot will start processing the flow-doc if any, otherwise expects voice_url. The execution will proceed as usual (other than expected full URL for all action URLs as mentioned above).


jamesaimonetti:
@icehess did you test creating a pivot callflow with both defined? Would make a good properly test to ensure we don't make changes to jesse/kz_json_schema that suddenly allow both to be defined.

icehess:
@jamesaimonetti Yes I did and it works:

- If both are not defined, you would get at least `not_one_schema_valid` error.
- If both are set you would get `more_than_one_schema_valid` error.

icehess:
Required PRs:
- master:
    - https://github.com/2600hz/kazoo-pivot/pull/52
    - https://github.com/2600hz/kazoo-callflow/pull/270
    - https://github.com/2600hz/kazoo-crossbar/pull/644
    - https://github.com/2600hz/kazoo-core/pull/1065
- 5.4:
    - https://github.com/2600hz/kazoo-pivot/pull/58
    - https://github.com/2600hz/kazoo-callflow/pull/284
    - https://github.com/2600hz/kazoo-crossbar/pull/670
    - https://github.com/2600hz/kazoo-core/pull/1100

PR #643, opened 2025-01-15
SUPP-565: descendants_count does not update after moving to another account

wdhammika:
Previously: The descendants count does not update after moving account into another account. 
Now: Added functionality to update descendants count for both previous parents and new parents.

Related PRs:
5.3- https://github.com/2600hz/kazoo-crossbar/pull/645

PR #645, opened 2025-01-17
[5.3] SUPP-565: descendants_count does not update after moving to another account

wdhammika:
Previously: The descendants count does not update after moving account into another account.

Related PRs :
master - https://github.com/2600hz/kazoo-crossbar/pull/643


wdhammika:
To maintain engineering standards about SUPP tickets policy, closing the PR to raise the same thing under KCRO domain.

PR #646, opened 2025-01-20
KCRO-234: descendants_count does not update after moving to another account

wdhammika:
Previously: The descendants count is not updating after moving the account into another account.

Related PRs
5.3 - https://github.com/2600hz/kazoo-crossbar/pull/647
5.4 - https://github.com/2600hz/kazoo-crossbar/pull/648


src/crossbar_util.erl
```diff
@@ -456,6 +456,8 @@ move_account(AccountId, AccountJObj, ToAccount, NewTree) ->
             lager:info("moved descendants of ~s", [AccountId]),
             _Services = kazoo_services_maintenance:update_tree(AccountId, NewTree, NewResellerId),
             lager:info("updating sevices tree for ~s", [AccountId]),
+            _ = descendants_count(lists:last(NewTree)),
```
nkarunarathne, line 459: Instead of just underscore, add meaningful variable names like "_Services "

wdhammika: The response of the descendants_count/1 is always 'ok'. Can update if it needs.
`-spec descendants_count(kz_term:proplist() | kz_term:ne_binary()) -> 'ok'.`

PR #657, opened 2025-02-14
fixing some invalid JSON schemas

icehess:
it's been long we didn't validate our JSON schemas, some adds invalid addtions. This fiexes them.


icehess:
core pr: https://github.com/2600hz/kazoo-core/pull/1081

PR #658, opened 2025-02-18
DUPT-566: fix load_user won't store db_doc

icehess:
The way the user doc is read is changed, without storing the original doc stored in db_doc causing any code that depends on that (welcome notification for example) to fail.

Also changing the code when webphone enabled notification is sent, to send it only when a webphone device is set or its current value is changed.


icehess:
PRs:
- master:
    - https://github.com/2600hz/kazoo-crossbar/pull/658
- 5.4
    - https://github.com/2600hz/kazoo-crossbar/pull/659

jamesaimonetti:
Is there a properly test that can be updated/added that captures the bad behaviour? or is it only tangential code affected (like the welcome email)?

icehess:
@jamesaimonetti you okay with merging this for now and have properly test later in other prs?

PR #680, opened 2025-04-01
KWEB-14: (Reliably) send account creation and deletion webhooks

artable:
Linked PRs:
https://github.com/2600hz/kazoo-teletype/pull/106
https://github.com/2600hz/kazoo-core/pull/1123
https://github.com/2600hz/kazoo-webhooks/pull/56

Crossbar previously had logic which determined if an account creation AMQP event would fire or not depending on the API request's <<"send_email_on_creation">> field.
Now crossbar will always send the event with a field for Teletype to determine if it sends an email or not. Webhooks now always receives these events, and can always send an `account_created` notification.

Crossbar now also sends `account_deleted` notifications, which Webhooks receives.

Object hooks currently can and will send `account_created` hooks. There is a race condition in creation, but this actually prevents a duplicate hook from being sent. Here's how that happens:
- An account_creation request happens
- The account doc in the account's database is created, triggering an object event
- Which Webhooks picks up. It searches in the `accounts` database for hooks, but the `accounts` doc hasn't been created yet, so webhooks doesn't find any hooks.
- the account doc in the `accounts` database is created, triggering an object event
- webhooks picks it up, and now that the `accounts` database exists, a hook is triggered.

Object hooks cannot send `account_deleted` hooks. The issue is that we hard delete the account doc in the `accounts` database when we delete the doc, so when Webhooks gets the event, it cannot find hooks. Some options I explored:
- `kapi_conf` deleted events do have an optional `doc` field which could include the just-deleted account doc, but by the time this event is fired from `kzs_doc`, the doc is long gone from memory.
- We could soft-delete the doc from the `accounts` database. This works just fine, though `kz_account_crawler` does find the soft-deleted doc and expects to run cron tasks on it, which fail and produce warnings. Other, more subtle side effects may exist.

Additionally from a user perspective it makes sense for account creation and deletion to be notification hooks. Account docs have a lot of data that is not particularly relevant to the end user. they're also very chatty and intrinsically require more inner-knowledge than notifications to set up correctly to avoid other object hooks making noise.

I also think it's intuitive. We have an account_created notification, so as an end user I would naturally assume we have an `account_deleted` notification. I think it's reasonable to argue the same about object hooks, but part of the idea is to push customers towards using the less advanced hook when possible.


icehess:
@artable There is a  lot of unrelated changes in this PR. I see commits are cherry-picked? Why?

Did you meant to rebase instead of cherry-picking? Remove/reset those commits, and rebase your branch on top of latest.

artable:
> @artable There is a lot of unrelated changes in this PR. I see commits are cherry-picked? Why?
> 
> Did you meant to rebase instead of cherry-picking? Remove/reset those commits, and rebase your branch on top of latest.

Looks like it's because of a pull I guess.

```bash
6743fb5f HEAD@{16}: pull (finish): returning to refs/heads/KWEB-14
6743fb5f HEAD@{17}: pull (pick): KZOO-589: Update docs with new value and remove deprecated values (#673)
c30a514f HEAD@{18}: pull (pick): KCAL-71 add property to callflow.user
66129f22 HEAD@{19}: pull (pick): KZOO-584: use access_token as auth token (#678)
16b68efa HEAD@{20}: pull (pick): KCRO-240: fix log message (#676)
e7c571a2 HEAD@{21}: pull (pick): KPIV-71: Add auto-record and min-required-recording-participants to conferences.profile.json (#663)
a9ac05c8 HEAD@{22}: pull (start): checkout a9ac05c8eb92f1402cd24acb8f4cef3b0207ce48```

File: src/modules/cb_accounts.erl [hunk 1]
```diff
@@ -979,11 +973,36 @@ notify_new_account(_Context, 'undefined') -> 'ok';
...
     _ = cb_context:put_reqid(Context),
     JObj = cb_context:doc(Context),
+    ShouldNotify = kz_term:is_true(cb_context:req_value(Context, <<"send_email_on_creation">>, 'true')),
+
     lager:debug("triggering new account notification for ~s", [cb_context:account_id(Context)]),
     Notify = [{<<"Account-Name">>, kzd_accounts:name(JObj)}
              ,{<<"Account-Realm">>, kzd_accounts:realm(JObj)}
              ,{<<"Account-API-Key">>, kzd_accounts:api_key(JObj)}
              ,{<<"Account-ID">>, cb_context:account_id(Context)}
+             ,{<<"Should-Notify">>, ShouldNotify}
```
icehess, line 1018:
```suggestion
             ,{<<"Should-Send-Email">>, ShouldNotify}
```

icehess (changes requested):

src/modules/cb_accounts.erl
```diff
@@ -979,11 +973,36 @@ notify_new_account(_Context, 'undefined') -> 'ok';
...
+-spec notify_deleted_account(cb_context:context(), kz_term:api_object()) -> 'ok'.
+notify_deleted_account(_Context, 'undefined') -> 'ok';
+notify_deleted_account(Context, _AuthDoc) ->
+    _ = cb_context:put_reqid(Context),
+    JObj = cb_context:doc(Context),
+
+    lager:debug("triggering account_deleted notification for ~s with parent ~s", [cb_context:account_id(Context), kzd_accounts:parent_account_id(JObj)]),
+    Notify = [{<<"Account-Name">>, kzd_accounts:name(JObj)}
+             ,{<<"Account-Realm">>, kzd_accounts:realm(JObj)}
+             ,{<<"Account-API-Key">>, kzd_accounts:api_key(JObj)}
```
icehess: Why sending api key?

icehess (changes requested):

src/modules/cb_accounts.erl
artable: The issue is Teletype does send other kinds of notifications too, right? It's an awkward property to name all around.

src/modules/cb_accounts.erl
artable: Why, because I copy and pasted the account-created notification here! What's not to understand?
...
...
...
Yeah uh that doesn't seem very necessary does it

artable:
CI failures are due to missing core.

PR #682, opened 2025-04-09
Adding a note to refresh/renew auth token doc

filipevarjao:
In the system_config/crossbar.auth we need to set a new value to "renew": true to default/auth_modules/cb_user_auth  e.g.:
```
...
 "cb_user_auth": {
        "renew": true,  <<---- this
        "enabled": true,
        "token_auth_expiry_s": 3600,
        "log_failed_attempts": true,
        "log_successful_attempts": true
      },
...
```
and flush configs.


jamesaimonetti:
I would have preferred a command included to show how to set this, both for the sysadmin and account admin, if appropriate

jamesaimonetti:
Also, how is the key path `foo/bar/baz`? lol, JSON key paths are typically `foo.bar.baz` and we really should include a fuller doc example like:
```json
{"default": {
    "auth_modules":{
        "cb_user_auth":{
            "renew":true
        }
     }
 }
}
```

PR #683, opened 2025-04-09
[fix-5.3.19] KCRO-234: update descendants count after moving accounts

wdhammika:
Previously: The descendants count is not updating after moving the account into another account.

Related PRs :
master - https://github.com/2600hz/kazoo-crossbar/pull/646
5.4 - https://github.com/2600hz/kazoo-crossbar/pull/648
5.3 - https://github.com/2600hz/kazoo-crossbar/pull/647


wdhammika:
Going to raise a new PR with disabling kazoo docs.

PR #693, opened 2025-04-17
KHEL-4: leak signup params as metadata

icehess:
Standardizing the private sign up fields, so each method or service that is sued for signing up the account can set some generic parameters and can be leaked as metadata from API when fetching account.

One of the parameter, `email_is_verified` needs further updates by other PRs to add the email verification steps, reminder, sending emails and etc.... The idea is that that the service or the method used for signup will have to provide those mechanism and set this flag true when email used for signup is verified. UI can use this generic flag to show a notice to user that the email is needed to be verified.

Also adding `pvt_is_trial` so and leak it via API, this is still wip, but currently we needed a way in CPAAS to determine an account is still in trial mode when using a feature we don't trail account have access. Other PRs will add the logic to update and remove this trial flag.

This trial flag is not time-based, at least for intended usage in CPAAS.


icehess:
PRs:

- master:
    - https://github.com/2600hz/kazoo-hello/pull/55
    - https://github.com/2600hz/kazoo-crossbar/pull/693
    - https://github.com/2600hz/kazoo-core/pull/1148

PR #704, opened 2025-05-14
KZOO-621: Remove enum checking of language

romanat-ooma:
As we are adding more and more languages with different TTS models, checking language here makes no sense anymore.

Depends on: https://github.com/2600hz/kazoo-core/pull/1174


jamesaimonetti:
@romanat-ooma is there a way to generate a doc snippet to insert into the media doc (or somewhere) that lists current voices expected to work? Or links to the various supported TTS services' languages?

priv/couchdb/schemas/media.json
```diff
@@ -95,113 +95,6 @@
                 "voice": {
                     "default": "female/en-US",
                     "description": "The voice to be used during the conversion",
-                    "enum": [
```
jamesaimonetti, line 98: can you add a minLength:1 or something?

jamesaimonetti: no empty strings allowed

romanat-ooma: fixed :)

PR #705, opened 2025-05-14
KCRO-246: require the 'flow' property on the callflow schema

jamesaimonetti:


jamesaimonetti:
Requires: https://github.com/2600hz/kazoo-properly/pull/242
Nice to have: https://github.com/2600hz/kazoo-callflow/pull/300

PR #719, opened 2025-06-02
KCRO-241: Inbound s/mms from Trunking IO

artable:
Linked PRs:
https://github.com/2600hz/kazoo-core/pull/1197

add `cb_inbound_messaging` for inbound s/mms hooks from kz_trunkingio.
specifically to handle TIO's unique im data format, especially hmac authentication

`cb_messaging` is generic, actual carrier-specific code is in [core](https://github.com/2600hz/kazoo-core/pull/1197).


src/modules/cb_messaging.erl
```diff
@@ -0,0 +1,360 @@
...
+-spec api_auth(kz_json:object(), cowboy:http_headers(), kz_term:api_ne_binary()) -> boolean().
+api_auth(_,_,_) -> 'false'.
+
+%%------------------------------------------------------------------------------
+%% @doc Given the path tokens related to this module, what HTTP methods are
+%% going to be responded to.
+%% @end
+%%------------------------------------------------------------------------------
+-spec allowed_methods() -> http_methods().
+allowed_methods() ->
```
icehess: No need for this function arity.

src/modules/cb_messaging.erl
```diff
@@ -0,0 +1,360 @@
...
+%%
+%% ```
+%%    /skels => []
+%%    /skels/foo => [<<"foo">>]
+%%    /skels/foo/bar => [<<"foo">>, <<"bar">>]
+%% '''
+%% @end
+%%------------------------------------------------------------------------------
+-spec resource_exists() -> boolean().
+resource_exists() -> 'false'.
```
icehess: No need for this function arity i guess.

src/modules/cb_messaging.erl
```diff
@@ -0,0 +1,360 @@
...
+            Ctx = set_data_in_context(Context, Body, QS),
+            {ok, Ctx};
+        {error, _} -> {error, parse_error}
+    end.
+
+get_request_json({ok, Body, _}) ->
+    decode_req_json(Body);
+get_request_json({error, _, _}) ->  {error, parse_error}.
+
+-spec decode_req_json(binary()) -> {ok, kz_json:object()} | {'error', 'parse_error'}.
```
icehess: Please add a comment here why we used a customized decode req json here. You will do your future-self a favor by doing this.

src/modules/cb_messaging.erl
```diff
@@ -0,0 +1,360 @@
...
+    Payload = cb_context:req_json(Context),
+    {Module, Func} = carrier_function('api_auth', Carrier),
+    Module:Func(Payload, cb_context:req_headers(Context), cb_context:account_id(Context)).
+
+-spec authorize(cb_context:context(), path_token()) -> boolean().
+authorize(_, _) -> 'true'.
+
+-type carrier_fun() :: 'api_auth' | 'im_type' | 'im_body' | 'request_ccvs' | 'request_kvs'.
+-spec carrier_function(carrier_fun(), kz_term:ne_binary()) -> {atom() | ?MODULE, carrier_fun()}.
+carrier_function(Function='api_auth', Carrier) ->
```
icehess: not really fan of this function tbh, remove it.

src/modules/cb_messaging.erl
```diff
@@ -0,0 +1,360 @@
...
+    case kz_module:is_exported(Module, Function, Arity) of
+        'true'->
+            {Module, Function};
+        'false' ->
+            lager:info("~s:~s/~p is not exported.", [Module, Function, Arity]),
+            {?MODULE, Function}
+    end.
+
+-spec im_type(kz_json:object()) -> 'sms' | 'mms'.
+im_type(_) -> 'sms'.
```
icehess: remove

src/modules/cb_messaging.erl
```diff
@@ -0,0 +1,360 @@
...
+
+-spec api_auth(kz_json:object(), cowboy:http_headers(), kz_term:api_ne_binary()) -> boolean().
+api_auth(_,_,_) -> 'false'.
```
icehess: remove

src/modules/cb_messaging.erl
```diff
@@ -0,0 +1,360 @@
...
+%%    /skels/foo/bar => [<<"foo">>, <<"bar">>]
+%% '''
+%% @end
+%%------------------------------------------------------------------------------
+-spec resource_exists() -> boolean().
+resource_exists() -> 'false'.
+
+-spec resource_exists(path_token()) -> boolean().
+resource_exists(?TRUNKING_IO) ->
+    'true'.
```
icehess:
```suggestion
resource_exists(Carrier) ->
    kz_term:module:ensure_loaded(<<"knm_", Carrier/binary>>).
```

icehess (changes requested):
Some notes. Plus:

- please remove this to `cb_inbound_messaging`
- document this new api endpoint :)) give sample, description and all


Also per our conversion:

- First of no really fan of `carrier_function`and `carrier_funs_exist`  functions. All you need is `kz_module:is_exported(<<"knm_", Carrier/binary>>).` in most case. Also not fan of that many callbacks from knm.
- use `put` instead of post.
 
Lets reduce amount of back and forth between cb and knm. Here is proposal:

- have cb call carrier `authenticate_inbound_message(AccountIdm Method, Url, Headers, QS, Body)` which simply return `'true' | `{error, ErrorMsg}` (method, qs, and url are options, you can drop them, at least pass account id, header and body)
- have cb calls knm carrier `validate_inbound_message` func. this simply returns the actual AMQP message to be sent if the pay is valid
    - if return ok, in crossbar side, run the extra validation, check for number, account id, set account if, standing and all.
    - then as last validator, call knm carrier `handle_mms_encoding`. This simply checks if the type is mms, handles fetching images and encode them.


I just realized that cb_mms and cb_sms are not actually storing message in database at all. i wasn't expecting this, but for now moving, just simplify the code and remove these clutter.

src/modules/cb_messaging.erl
artable: My comment is on the callback, `request_data`
```erlang
%% @doc Requests from TrunkingIO include null keys, which crossbar_util
%% implicitly filters out when using kz_json:set_keys(). HMAC auth requires
%% the payload as sent, including null kvs.
%% @end
```

romanat-ooma (approved):

doc/inbound_messaging.md
```diff
@@ -0,0 +1,45 @@
...
+See [TIO API Docs](https://gitlab.com/oomaforbin/onsip/trunkingio-sms/-/blob/main/docs/api/messages.md?ref_type=heads) for expected schema.
+
+## Schema
+
+
+
+
+## Examples
+
+> PUT /hooks/accounts/{ACCOUNT_ID}/messaging/{CARRIER}
```
jamesaimonetti: should the URI be `/inbound_messaging` to match the cb module name and the bindings it is using?

src/modules/cb_inbound_messaging.erl
```diff
@@ -0,0 +1,328 @@
...
+
+%%------------------------------------------------------------------------------
+%% @doc Authenticates the incoming request, returning true if the requestor is
+%% known, or false if not.
+%% @end
+%%------------------------------------------------------------------------------
+-spec authenticate(cb_context:context(), path_token()) -> boolean().
+authenticate(Context, Carrier) ->
+    Payload = cb_context:req_json(Context),
+    KzdCarrier = binary_to_atom(<<"knm_", Carrier/binary>>),
```
jamesaimonetti: use `kz_term:to_*`

src/modules/cb_inbound_messaging.erl
```diff
@@ -0,0 +1,328 @@
...
+    case kz_module:is_exported(KzdCarrier, 'authenticate_req', 3) of
+        true ->
```
jamesaimonetti: tickie the atomzzz

doc/inbound_messaging.md
jamesaimonetti: Also PUT here but POST below?

src/modules/cb_inbound_messaging.erl
```diff
@@ -0,0 +1,328 @@
...
+authorize(_, _) -> 'true'.
+
+%%------------------------------------------------------------------------------
+%% @doc Given the path tokens related to this module, what HTTP methods are
+%% going to be responded to.
+%% @end
+%%------------------------------------------------------------------------------
+-spec allowed_methods(path_token()) -> http_methods().
+allowed_methods(_Carrier) ->
+    [?HTTP_POST].
```
jamesaimonetti, line 134: Should this be PUT?

src/modules/cb_inbound_messaging.erl
```diff
@@ -0,0 +1,328 @@
...
+resource_exists(?TRUNKING_IO) ->
+    kz_module:ensure_loaded(knm_trunkingio) =/= false
+        andalso kz_module:ensure_loaded(kzd_trunkingio_im) =/= false;
+resource_exists(_) -> false.
+
+-spec put(cb_context:context(), path_token()) -> cb_context:context().
+put(Context, _Carrier) ->
+    Payload = cb_context:doc(Context),
+    _ = kz_amqp_worker:cast(Payload, fun kapi_im:publish_inbound/1),
+    crossbar_util:response(kz_json:normalize(kz_api:remove_defaults(Payload)), Context).
```
jamesaimonetti, line 163: Would it make sense to return a 202 since we don't know if the AMQP payload is actually processed successfully?

src/modules/cb_inbound_messaging.erl
```diff
@@ -0,0 +1,328 @@
...
+    create(Context, Carrier).
+
+%%------------------------------------------------------------------------------
+%% @doc Create a new instance with the data provided, if it is valid
+%% @end
+%%------------------------------------------------------------------------------
+-spec create(cb_context:context(), path_token()) -> cb_context:context().
+create(Context, Carrier) ->
+    OnSuccess = fun(Ctx) -> on_successful_validation(Ctx, Carrier) end,
+    KzdCarrier = binary_to_atom(<<"kzd_", Carrier/binary, "_im">>),
```
jamesaimonetti: `kz_term:to_*`

src/modules/cb_inbound_messaging.erl
```diff
@@ -0,0 +1,328 @@
...
+
+handle_carrier_validate(Context, {error, _Error}) ->
+    cb_context:add_system_error('parse_error', Context);
+handle_carrier_validate(Context, {ok, JObj}) ->
+    cb_context:set_doc(Context, JObj).
+
+-spec to_number_exists(cb_context:context()) -> cb_context:context().
+to_number_exists(Context) ->
+    Doc = cb_context:doc(Context),
+    [Number] = kz_im:to(Doc),
```
jamesaimonetti: What enforces 1 and only 1 element in the list?

src/modules/cb_inbound_messaging.erl
```diff
@@ -0,0 +1,328 @@
...
+-spec account_is_in_good_standing(cb_context:context()) -> cb_context:context().
+account_is_in_good_standing(Context) ->
+    AccountId = cb_context:account_id(Context),
+    case kz_services_standing:acceptable(AccountId) of
+        {'true', _} -> Context;
+        {'false', #{message := Msg}} -> cb_context:add_system_error('account', Msg, Context)
+    end.
+
+-spec account_has_messaging_enabled(cb_context:context()) -> cb_context:context().
+account_has_messaging_enabled(Context) ->
```
jamesaimonetti, line 308: I would like to see more functions used. For instance:

```erlang
account_has_messaging_enabled(Context) ->
    account_has_messaging_enabled(Context, kz_im:type(cb_context:doc(Context))).
    %% since you don't use Doc later, you can just access the field you want and check in next function

account_has_messaging_enabled(Context, <<"sms">>) ->
    case is_sms_enabled() of
        ....
    end;
account_has_messaging_enabled(Context, <<"mms">>) -> ...
```

src/modules/cb_inbound_messaging.erl
```diff
@@ -0,0 +1,328 @@
...
+                         ),
+            cb_context:add_system_error('account', <<"service temporarily unavailable">>, Context)
+    end.
+
+-spec reseller_has_messaging_enabled(cb_context:context()) -> cb_context:context().
+reseller_has_messaging_enabled(Context) ->
+    Doc = cb_context:doc(Context),
+    case kz_im:type(Doc) of
+        <<"sms">> ->
+            case kz_services_im:is_sms_enabled(cb_context:reseller_id(Context)) of
```
jamesaimonetti: You could extend account_has_messaging_enabled to call `has_messaging_enabled(Context, cb_context:account_id(Context))` and `reseller_has_messaging_enabled` calls `has_messaging_enabled(Context, cb_context:reseller_id(Context))` and then do the `has_messaging_enabled(Context, AccountId) -> has_messaging_enabled(Context, AccountId, kz_im:type(cb_context:doc(Context)).` etc

jamesaimonetti (changes requested):

src/modules/cb_inbound_messaging.erl
```diff
@@ -0,0 +1,328 @@
...
+        {error, _} -> {error, parse_error}
+    end.
+
+get_request_json({ok, Body, _}) ->
+    decode_req_json(Body);
+get_request_json({error, _, _}) ->  {error, parse_error}.
+
+-spec decode_req_json(binary()) -> {ok, kz_json:object()} | {'error', 'parse_error'}.
+decode_req_json(Body) ->
+    try kz_json:unsafe_decode(Body) of
```
jamesaimonetti: why are you recreating `kz_json:decode` here? Would not `kz_json:decode(Body, [{default, 'undefined'}])` suffice? and pattern match on 'undefined' above?

src/modules/cb_inbound_messaging.erl
artable: Well. Yes. I asked this very question.
We're not doing that right now though--the tools that make that kind of... correct error reporting valuable are kicked down the road on the TIO side too.

src/modules/cb_inbound_messaging.erl
artable: That's a good point.
TIO only ever sends one item in the list, if there are multiple recipients it uses the 'other recipients' field. TIO sends separate hooks for each recipient though. 

If they want to change that it'll be a major code change and should fail until fixed, I'll enforce this as part of validation.

src/modules/cb_inbound_messaging.erl
artable: Being fully honest I knew you were going to say this if and when you saw the PR. Shame on me!

src/modules/cb_inbound_messaging.erl
icehess: Not following why it is a major change? here you only need to loop over the list, no?

src/modules/cb_inbound_messaging.erl
artable: Mimicing the behavior of `api_util`s normal processing as much as is possible. So... ask [_checks git blame_](https://github.com/2600hz/kazoo-crossbar/blob/db658602ad838c705ca34dd8277f14f287f87c00/src/api_util.erl#L690) you, karl, Mark, Daniel, and Luis.

Listen I wasn't gonna mess with a function touched 5+ years ago by those five people.

src/modules/cb_inbound_messaging.erl
artable: It'd be a larger change on their end than ours, but we have to reflect it in a fair number of ways.
Biggest problem I see is if not all the `to` numbers exist in kazoo or if they can't be routed for other reasons. How do we want to handle a partial success?

artable:
There's a security flaw in this module:
a reseller account can route messages to another resellers child numbers. We authorize them based on the reseller in the URL, but never check to make sure that the number belongs to an account under that reseller.

src/modules/cb_inbound_messaging.erl
artable: Further issues:
if one number is invalid, `cb_sms` just fails the whole request. It's not that simple here.
cb_sms has that luxury because it's expected all the numbers are from the same account. If one number fails because there's insufficient funds, all of them do. Fail the request.
With tio a single valid hook could span multiple end-user accounts, each with different funding statuses. If TIO sends a list of `to` numbers for one reseller across five end-user accounts, two of which have sms disabled, one which has insufficient funds, and two which are in good standing and can do sms, how should we respond?
we shouldn't fail the whole request, two 'innocent' accounts wouldn't get messages because a sibling account isn't in good standing. At a minimum we'd need to turn the whole validation structure upside down to deal with that.

This also makes the 'different reseller's numbers' thing far more complicated on TIO's side. But that's their problem.

I could see a world where we do get to fail the whole request, but we'd still need to compose a response that lets tio remove offending numbers and try again, and that also requires turning the validation structure upside down to deal with it.

So skip it until someone wants it.

jamesaimonetti: my opinion is to fail the request with the failing numbers, as you suggest. If TIO wants to offer a request with bogus data, regardless of valid data that may also be contained therein, KAZOO should reject the request. We don't accept a request with schema validation errors because the other fields are passing fine.

src/modules/cb_inbound_messaging.erl
jamesaimonetti: but you aren't returning the JSON error to the client so the two use cases aren't the same. here you just check for error, log it, and return generic `parse_error` with no instructions to the client on where parsing went wrong

src/modules/cb_inbound_messaging.erl
```diff
@@ -0,0 +1,369 @@
...
+%%------------------------------------------------------------------------------
+%% @doc Requests from TrunkingIO include null keys, which crossbar_util
+%% implicitly filters out when using kz_json:set_keys(). HMAC auth requires
+%% the payload as sent, including null kvs.
+%% Note that req_data and req_json are set differently-- req_data is
+%% the full unmodified original request, req_json is set with the usual
+%% key normalization.
+%% @end
+%%------------------------------------------------------------------------------
+-spec request_data(cb_context:context(), path_token()) -> {'ok', cb_context:context()} |
```
jamesaimonetti: first arg's type does not match function's arg

src/modules/cb_inbound_messaging.erl
```diff
@@ -0,0 +1,369 @@
...
+        'true' ->
+            Module = kz_term:to_atom(CarrierMod),
+            case Module:authenticate_req(Payload, cb_context:req_headers(Context), cb_context:account_id(Context)) of
+                'true' -> 'true';
+                'false' -> {'stop', cb_context:add_system_error('invalid_credentials', Context)}
+            end;
+        _ -> {'stop', cb_context:add_system_error('invalid_credentials', Context)}
+    end.
+
+to_carrier_mod(Carrier) -> <<"knm_", Carrier>>.
```
jamesaimonetti: you are missing `/binary` after Carrier; otherwise Erlang will expect Carrier to be `integer()`

jamesaimonetti: also suggests this hasn't been tested yet, as this would crash if a request was sent in to test this

jamesaimonetti (changes requested):

src/modules/cb_inbound_messaging.erl
icehess: the code is updated now, now it tracks the account id of the first number, if it changes will fail with proper validation error.

As a reminder, the initial account id on the URL is expected to be the reseller of the number. This is how tio webhook is created by kazoo-trunkingio app.

This endpoint should lookup the account of the first number and make sure the reseller is is in hierarchy then check of all numbers are belong to the same account. multiple number across different account is not supported.

Also we checked with tio, they indeed send it a single number array, even if it is a group message, it fires one webhook per number. This code was causing issues previously, but since it is new code, and we just try to sanity check stuff.

src/modules/cb_inbound_messaging.erl
icehess: updated, i think kz_json already logs on error case. we just need tpg log success decode just like api_utils

src/modules/cb_inbound_messaging.erl
icehess: i can't type apparently :sad_face:

PR #746, opened 2025-07-14
KZOO-568 add address doc and API

jingxu2021:
Add cb_addresses.erl for handling address docs.


src/modules/cb_addresses.erl
```diff
@@ -0,0 +1,244 @@
...
+
+%%------------------------------------------------------------------------------
+%% @doc If the HTTP verb is DELETE, execute the actual action, usually a db delete
+%% @end
+%%------------------------------------------------------------------------------
+
+-spec delete(cb_context:context(), path_token()) -> cb_context:context().
+delete(Context, Id) ->
+    Numbers = used_by_numbers(Context, Id),
+    case length(Numbers) of
```
jamesaimonetti: Don't call `length(List)` if you don't have to, pattern-matching an empty list is right there:

```erlang
case used_by_numbers(Context, Id) of
    [] -> delete;
    Numbers -> ...
end
```

src/modules/cb_addresses.erl
```diff
@@ -0,0 +1,244 @@
...
+                                            kz_json:from_list([{<<"Reason">>, <<"Address is used by these numbers">>},
+                                                               {<<"Numbers">>, Numbers}])),
+            cb_context:set_resp_error_code(Ctx1, 409)
+    end.
+
+used_by_numbers(Context, Id) ->
+    Option= [{'key', Id}],
+    AccountDb = cb_context:db_name(Context),
+    case kz_datamgr:get_results(AccountDb, ?NUM_LIST, Option) of
+        {'ok', Numbers} -> lists:map(fun(E) -> kz_json:get_binary_value(<<"id">>, E) end, Numbers);
```
jamesaimonetti: `[kz_doc:id(E) || E <- Numbers]` in general (maybe not the case as much anymore) list comprehensions are more compact for simple mappings like this

src/modules/cb_addresses.erl
```diff
@@ -0,0 +1,244 @@
...
+maybe_create_view(Context, _OtherFailure) ->
+    Context.
+
+%%------------------------------------------------------------------------------
+%% @doc Normalizes the results of a view
+%% @end
+%%------------------------------------------------------------------------------
+-spec normalize_view_results(kz_json:object(), kz_json:objects()) -> kz_json:objects().
+normalize_view_results(JObj, Acc) ->
+    Ret0 = kz_json:get_value(<<"value">>, JObj),
```
jamesaimonetti: Use typed getter (I assume `kz_json:get_json_value/2`) and no need to create `Ret0` I don't think, just `[kz_json:get_json_value(<<"value">>, JObj) | Acc]` is short enough.

jamesaimonetti (changes requested):
First pass

src/modules/cb_addresses.erl
jingxu2021: Fixed, thanks

src/modules/cb_addresses.erl
jingxu2021: Fixed, thanks

src/modules/cb_addresses.erl
jingxu2021: Fixed, thanks

src/modules/cb_addresses.erl
```diff
@@ -0,0 +1,242 @@
...
+        ,put/1
+        ,post/2
+        ,patch/2
+        ,delete/2
+        ]).
+
+-include("crossbar.hrl").
+
+-define(NUM_LIST, <<"addresses/used_by_numbers">>).
+-define(CB_LIST, <<"addresses/list_by_id">>).
```
jamesaimonetti, line 29: see comment on core PR about using crossbar_listings design doc

src/modules/cb_addresses.erl
```diff
@@ -0,0 +1,242 @@
...
+patch(Context, Id) ->
+    post(Context, Id).
+
+%%------------------------------------------------------------------------------
+%% @doc If the HTTP verb is DELETE, execute the actual action, usually a db delete
+%% @end
+%%------------------------------------------------------------------------------
+
+-spec delete(cb_context:context(), path_token()) -> cb_context:context().
+delete(Context, Id) ->
```
jamesaimonetti, line 179: Is `Id` the AddressId ? If so, can we use `AddressId` throughout for readability?

jamesaimonetti (changes requested):

